PR #54, opened 2020-02-17
Add real ships from MarineTraffic

haakov:
Poll MarineTraffic every 2 minutes for updated ship positions in the Juan de Fuca strait area.

The API key should expire some time during Tuesday (Feb 18).


haakov:
The new API key is now live, it will work for 3 days.

tylerlum:
Simon said we're getting access to AIS data in March :) will be great to use this at that time

haakov:
Yeah! I suppose it will have to be handled differently, being a different service and all, but I believe we're already using pynmea (?) somewhere it shouldn't be too much work

tylerlum:
Closing, as replaced by #225 and #192

PR #76, opened 2020-04-05
Try straight path

tylerlum:
The purpose of this PR is to add functionality to bias towards simple straight lines unless they do not work. This is done with ompl's simplifySolution() function. This method removes all special kinks in a path. It also loses proper tacking behavior if it is desired.

This is done by:
* If no valid solutions can be found in the pathfinding time limit, it will not do any simplifying. 
* If valid solutions exist, it will find the minimum cost between all unsimplified solutions and simplified solutions. Then returns the best one :)

### DIFF
--- a/python/utilities.py
+++ b/python/utilities.py
@@ -176,7 +176,6 @@ def createLocalPathSS(state, runtimeSeconds=2, numRuns=4, plot=False):
             invalidSolutions.append(solution)
 
     # If no solutions found, re-run with larger runtime
-    # TODO: Figure out if there is a better method to handle this case
     totalRuntimeSeconds = numRuns * runtimeSeconds
     while len(solutions) == 0:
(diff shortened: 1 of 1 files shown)


python/utilities.py
```diff
@@ -187,44 +188,74 @@ def isValidSolution(solution, referenceLatlon, state):
...
+            # Check unsimplified path
+            unsimplifiedPath = og.PathGeometric(solution.getSolutionPath())
+            unsimplifiedCost = unsimplifiedPath.cost(solution.getOptimizationObjective()).value()
+            if unsimplifiedCost < minCost:
+                bestSolution = solution
+                bestSolutionPath = unsimplifiedPath
+                minCost = unsimplifiedCost
+
+            # Check simplified path
+            solution.simplifySolution()
```
tylerlum, line 219: @patrick-5546 this is the simplifying thing. Should probably check when this part was removed and why.

PR #126, opened 2020-06-19
Update obstacle rep

tylerlum:
The purpose of this PR is to update the obstacle rep to allow the AIS boats to be projected forward to where they would be at the closest point of collision.

The main changes are:

* Separate ObstacleInterface and ShapeInterface so that each obstacle type is separate from implementation. This allows us to visualize both the actual current AIS boat position and the projected position. The projected position is fully opaque blue and the current position is transparent red.
* Reorganizing the code to be more clear and to avoid circular dependencies (eg. moving getObstacles to obstacles.py and createPath to Path.py). Don't want utilities import Path and Path import utilities, as this is not good practice.
* Fix unit tests to work with new obstacle rep

This can be tested with:

`roslaunch local_pathfinding all.launch`, then `rosparam set /obstacle_type ellipse` (replace ellipse with circles or wedge or hyrbrid_ellipse or hybrid_circle) or `rostopic pub /boat_on_path local_pathfinding/addBoat "addType: 'trailing'
ship: {ID: 0, lat: 0.0, lon: 0.0, headingDegrees: 0.0, speedKmph: 1.0}
waypointIndex: 0"`

This is a pretty huge PR, so this may cause bugs. I have tested this a good amount to be okay merging it, but we should be looking out for bugs. Please try it out and see how it works!


haakov (approved):
Huge change, but looks good to me! Tested, works well

PR #139, opened 2020-10-07
added launch argument to change number of mock AIS boats

JoshuaJSam:
Hmm, I made a mistake in the file commits and accidentally pushed my entire install folder :\   . During the meeting If someone can help me with git, can you show me how I fix this other than making a new branch from master and re-adding my changes there?  Also can you explain what the circleci lint failure means? Otherwise the launch argument has been added and is working as expected.

PR #140, opened 2020-10-09
added launch argument to change number of mock AIS boats

JoshuaJSam:


JoshuaJSam:
@tylerlum I'm not sure why but my original PR at https://github.com/UBCSailbot/local-pathfinding/pull/139 isn't syncing with the latest commit. I made this new PR here

python/MOCK_AIS.py
```diff
@@ -12,7 +12,7 @@
 
 # Constants
 AIS_PUBLISH_PERIOD_SECONDS = 0.1  # Keep below 1.0 for smoother boat motion
-NUM_AIS_SHIPS = 30
+NUM_AIS_SHIPS = rospy.get_param('num_ais_ships', '5')
```
tylerlum, line 15: I think the default value should be `5` instead of `'5'` so that it will be an int, not a string

launch/launch_all_mocks.launch
```diff
@@ -24,6 +24,10 @@
     <arg name="obstacle_type" default="hybrid_circle" />
     <param name="obstacle_type" type="string" value="$(arg obstacle_type)" />
 
+    <!-- Sets the num_ais_ships -->
```
tylerlum, line 27: This is perfect

launch/all.launch
```diff
@@ -10,6 +10,7 @@
     <arg name="gps_file" default="" />
     <arg name="goal_file" default="" />
     <arg name="wind_file" default="" />
+    <arg name="num_ais_ships" default="5" />
```
tylerlum, line 13: This is good, but you need to add something like
```
      <arg name="num_ais_ships" value="$(arg num_ais_ships)"/>
```
in the launch_all_mocks.launch tag

tylerlum (changes requested):
Looks good! I tried it out and it didn't quite work, I think you missed one key line in the all.launch. Also, you should run `git merge master` from this branch to make to align with the newest updates.

Issue #129, opened 2020-09-05
Add summarization of a run


tylerlum:
Right now, we can run a simulation and see the results. We have lots of logging being done, which is great. However, we should try to get a summary of important takeaways based on each run:

Eg. Total time. Total Distance covered. # of bad paths. Worst path created. Closest we got to an obstacle. Plot of distance to closest obstacle over the whole run. Number of collisions. Number of times pathfinding couldn't find a path in time.

This can be shown visually or with a text file or even printed to the screen at the end.

tylerlum:
Breaking this into a few specific different tasks. Note: for an example of an independent ros node, look at https://github.com/UBCSailbot/local-pathfinding/blob/master/python/addBoat.py

1. Create an independent ros node that subscribes to the sailbot GPS and AIS boat locations. Then finds which AIS boat is closest to our boat, calculate that distance, and then store the distance and time in a text file. Repeat this every second (can make a constant UPDATE_TIME_SECONDS = 1.0 that can be modified). When the simulation ends, we will have a .txt file (or something similar) that allows us to plot distance vs. time through the run. Perhaps even save a plot of this, rather than saving to a text file.

2. Create an independent ros node that subscribes to the sailbot GPS. Every second (can make a constant UPDATE_TIME_SECONDS = 1.0 that can be modified), store the sailbot GPS. Then at the end, plot the distance traveled over time OR displacement from initial position over time OR plot the full path that the boat took (see how jagged or uneven it is).

3. Go to https://github.com/UBCSailbot/local-pathfinding/blob/master/python/Path.py#L774

tylerlum:
@haakov @JoshuaJSam @patrick-5546 please comment here which part you plan to work on once you decide so you don't double up on the same task

patrick-5546:
I can do part 5

JoshuaJSam:
Part 2 looks interesting! I'd like to get some experience working on my own ROS node :)

haakov:
I'm doing 1 and 3

tylerlum:
I'm doing number 4

patrick-5546:
I'll do 1

patrick-5546:
I'll do 3

patrick-5546:
All 5 tasks complete :)

PR #147, opened 2020-10-24
Log collisions

patrick-5546:
Revised collision_checker so it only prints collision and warning messages once, counts the number of collisions and takes a screenshot when each collision/warning occurs. Not sure if I reset main_loop correctly, I should have only changed all.launch and collision_checker in this pr.

Task 5 of #129


patrick-5546:
After I pullled from master, I noticed that some new warnings were overloading the terminal, so it is harder to find the warnings from collision checker (if that matters).

launch/all.launch
```diff
@@ -31,5 +31,5 @@
 
     <node pkg="local_pathfinding" type="main_loop.py" name="main_loop" output="$(arg main_loop_output)"/>
     <node pkg="local_pathfinding" type="local_path_visualizer.py" name="local_path_visualizer" />
-    <node pkg="local_pathfinding" type="collision_checker.py" name="collision_checker" />
+    <node pkg="local_pathfinding" type="collision_checker.py" name="collision_checker" output="screen" />
```
tylerlum, line 34: Perfect

python/collision_checker.py
```diff
@@ -27,22 +30,51 @@ def GPS_callback(self, data):
...
+            for ship in self.ships:
+                dist = distance((ship.lat, ship.lon), (self.lat, self.lon)).km
+
+                if dist < self.collision_radius:
+                    check_ship = ship
+                    self.times_collided = self.times_collided + 1
+                    rospy.logfatal("Boat has collided with obstacle. Collision radius {}km.\n"
+                                   "\tActual distance to boat: {}km. Collision number: {}"
+                                   .format(self.collision_radius, dist, self.times_collided))
+                    utilities.takeScreenshot()
```
tylerlum, line 46: Great use of this :D

python/collision_checker.py
```diff
@@ -27,22 +30,51 @@ def GPS_callback(self, data):
...
+
     rate = rospy.Rate(1 / CHECK_PERIOD)
 
+    shipColliding = None
     while not rospy.is_shutdown():
-        collision_checker.check_for_collisions()
+        shipColliding = collision_checker.check_for_collisions(shipColliding)
         rate.sleep()
+
+    print("Number of collisions: {}. Number of warnings: {}."
```
tylerlum, line 79: Glad that this works now

python/collision_checker.py
```diff
@@ -27,22 +30,51 @@ def GPS_callback(self, data):
...
+    def get_times_warned(self):
+        return self.times_warned
 
 
 if __name__ == '__main__':
     collision_checker = CollisionChecker(COLLISION_RADIUS_KM, WARN_RADIUS_KM)
+
     rate = rospy.Rate(1 / CHECK_PERIOD)
 
+    shipColliding = None
```
tylerlum, line 74: This solution you implemented definitely looks like it works, but I feel that it could be updated to be a bit more clear.

python/collision_checker.py
```diff
@@ -27,22 +30,51 @@ def GPS_callback(self, data):
...
-                rospy.logfatal("Boat has collided with obstacle. Collision radius {}km. Actual distance to boat: {}km"
-                               .format(self.collision_radius, dist))
+    def check_for_collisions(self, colliding_ship):
+        check_ship = colliding_ship
 
-            elif dist < self.warn_radius:
-                rospy.logwarn("Close to collision. Within {}km of boat. Actual distance to boat: {}km"
-                              .format(self.warn_radius, dist))
+        if check_ship is None:
+            for ship in self.ships:
```
tylerlum, line 37: This for loop does not have any breaks. If it did happen to be close to 2 different boats, it would always return the second one. Then later you have a "Other ships are not within the collision or warning radius anymore" message if that second boat leaves the radius, but there could still be the first boat that is still within the radius.

python/collision_checker.py
tylerlum, line 74: I think the logic could be improved by removing the use of shipColliding and replacing it with a class variable liist/set in CollisionChecker. Eg.`self.currentlyCollidingBoats` and `self.currentlyWarningBoats`.

Then in your for loop looking at boats, you can check if the boat is in `self.currentlyCollidingBoats` or `self.currentlyWarningBoats`.

If it is in that list, then it is the same boat. No new collision. If it is a new boat, then add it the list and increment the counter. You can also check the number of boats in `self.currentlyCollidingBoats` or `self.currentlyWarningBoats` that were seen in the for loop.

Let me know if that makes sense or you feel something else would be better. This is just my suggestion on how to ensure this code has behavior we want.

tylerlum (changes requested):

python/collision_checker.py
tylerlum: You may also double count the collision of the first boat in this case

.gitignore
```diff
@@ -84,4 +84,6 @@ install/castxml/
 *.out
 *.app
 
-
+# Screenshots
+*images/*
```
tylerlum, line 88: Awesome!

python/collision_checker.py
```diff
@@ -28,21 +33,44 @@ def AIS_callback(self, data):
...
         for ship in self.ships:
             dist = distance((ship.lat, ship.lon), (self.lat, self.lon)).km
-            if dist < self.collision_radius:
-                rospy.logfatal("Boat has collided with obstacle. Collision radius {}km. Actual distance to boat: {}km"
-                               .format(self.collision_radius, dist))
 
-            elif dist < self.warn_radius:
-                rospy.logwarn("Close to collision. Within {}km of boat. Actual distance to boat: {}km"
-                              .format(self.warn_radius, dist))
+            if dist < max_radius and dist >= min_radius:
```
tylerlum, line 46: Try doing `min_radius <= dist < max_radius`

python/addBoat.py
```diff
@@ -14,23 +14,30 @@
 
 
 def command_callback(msg):
+    ships = rospy.wait_for_message('/AIS', AISMsg).ships
+    newShipID = 0
```
tylerlum, line 18: This is great and definitely works, but you can try this more "pythonic" way

`newShipID = max([ship.ID for ship in ships]) + 1`

tylerlum (approved):
Great work! Consider adding the nice python things

PR #148, opened 2020-10-28
Store paths

tylerlum:
The purpose of this PR is to add an independent ROS node that:
* Stores local paths to a pickle file
* Creates a plot showing the path cost over time. This includes the breakdown of the cost

cost_plot
cost_breakdown_plot


JoshuaJSam (approved):

patrick-5546 (commented):
cool

PR #156, opened 2020-11-06
Log closest obstacle

patrick-5546:
Finishing task 1 of #129

### DIFF
--- a/python/log_closest_obstacle.py
+++ b/python/log_closest_obstacle.py
@@ -10,6 +10,11 @@ import matplotlib.pyplot as plt
 import numpy as np
 
 UPDATE_TIME_SECONDS = 1.0
+PLOT_RANGE = 5
+
+# match with collision_checker.py
+COLLISION_RADIUS_KM = 0.1
+WARN_RADIUS_KM = 0.3
 
 # Output paths
 dateStr = date.today().strftime("%b-%d-%Y")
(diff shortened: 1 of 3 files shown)


python/log_closest_obstacle.py
```diff
@@ -0,0 +1,80 @@
...
+        self.ships = data.ships
+
+    def GPS_callback(self, data):
+        self.lat = data.lat
+        self.lon = data.lon
+
+    def find_closest_ship(self):
+        closest_ship_dist = None if len(self.ships) == 0 else min([distance((ship.lat, ship.lon),
+                                                                            (self.lat, self.lon)).km
+                                                                  for ship in self.ships])
```
patrick-5546, line 48: This might be hard to understand, I just wanted to try out the list comprehension thing. If you want, I can change it back to a normal for loop.

patrick-5546: Or maybe I could make a distance function?

patrick-5546: Added distance function

python/log_closest_obstacle.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def GPS_callback(self, data):
+        self.lat = data.lat
+        self.lon = data.lon
+
+    def get_distance(self, ship):
+        return distance((ship.lat, ship.lon), (self.lat, self.lon)).km
+
+    def find_closest_ship(self):
+        closest_ship_dist = None if len(self.ships) == 0 else min([self.get_distance(ship) for ship in self.ships])
```
tylerlum, line 48: This is beautiful :')

python/log_closest_obstacle.py
```diff
@@ -0,0 +1,80 @@
...
+timeStr = datetime.now().strftime('%H-%M-%S')
+ABS_PATH_TO_THIS_FILE_DIR = os.path.dirname(os.path.abspath(__file__))
+ABS_PATH_TO_OUTPUT_DIR = os.path.join(ABS_PATH_TO_THIS_FILE_DIR, "../output/{}_{}".format(dateStr, timeStr))
+ABS_PATH_TO_OUTPUT_MIN_DISTANCE_FILE = os.path.join(ABS_PATH_TO_OUTPUT_DIR, "minDistance.pkl")
+ABS_PATH_TO_OUTPUT_MIN_DISTANCE_PLOT_FILE = os.path.join(ABS_PATH_TO_OUTPUT_DIR, "min_distance_plot.png")
+
+
+class LogClosestObstacle:
+    def __init__(self):
+        rospy.init_node('log_closest_obstacle', anonymous=True)
```
tylerlum, line 25: Great work, super clear and clean code

tylerlum (approved):
image

Just tried this out, works beautifully!

tylerlum:
min_distance_plot

Just ran this overnight, this looks awesome. I think it would actually be cool if you had a horizontal line for the WARN RADIUS and COLLISION RADIUS distances to see what is normal, and/or making the y axis upper limit be something like 10km, so we get more detail about the close scenarios.

tylerlum:
If closest boat is farther than 10km away, that's good and we don't really need to know exactly the number.

patrick-5546:
Implemented your suggestions, with max distance on graph being 5 km (can easily be changed).

min_distance_plot

PR #157, opened 2020-11-07
Implemented functionality

patrick-5546:
Task 3 of #129


tylerlum (approved):
Fix lint error then good

Issue #158, opened 2020-11-13
Store file to create dynamic plot (zoom and pan)


tylerlum:
I've noticed that the longer we run the plots for, the harder it is to take away insights since we can't zoom in very well. Take a look at this https://stackoverflow.com/questions/4348733/saving-interactive-matplotlib-figures might be good to just save the pickle file of the figure, then have a new python script to visualize the figures, allowing you to change the scale and range of the plot to look more closely at certain areas 

```
And there's gcf and gca (get current figure/axis). Same as in MATLAB if my memory is correct.
In [68]: plt.gca()
Out[68]: <matplotlib.axes._subplots.AxesSubplot at 0xb17a35cc>
In [66]: plt.gcf()
Out[66]: <matplotlib.figure.Figure at 0xb091eeec>
```

tylerlum:
Done in #195 mostly

PR #160, opened 2020-11-14
Revised minDistance plot to be interactive

patrick-5546:
Addresses #158 

Change the output folder name at the bottom to choose which run you want to plot.
Will plot a closeup of each time there is an obstacle within the warning radius. 
Figure 1 is the overall graph, and figure 2 is the first closeup. After closing both of them, the next closeup will appear.

Screenshot from 2020-11-14 13-13-04


python/interactive_graph.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # Getting warning times
+    i = 1  # i = 0 is y intercept, not wanted
+    warn_start_times = list()
+    while i < len(y):
+        if y[i] < WARN_RADIUS_KM:
+            warn_start_times.append(x[i])
+            while i + 1 < len(y) and y[i + 1] >= y[i]:
+                i += 1
+        i += 1
```
patrick-5546, line 27: Is there a better way of doing this? Seems pretty inefficient.

tylerlum: Yeah good point. Probably not the biggest deal, but probably could have something like:

for i in range(1, len(y)):
    if y[i] < WARN and y[i-1] > WARN:
        warn_start_times.appned(i)

python/interactive_graph.py
```diff
@@ -0,0 +1,40 @@
...
+    # Sequentially looking around each warning start time
+    for warning in warn_start_times:
+        pl.load(open(PATH_TO_PKL, 'rb'))
+        print ("Warning around {} seconds" .format(warning))
+        plt.ylim(0, 1)
+        plt.xlim(warning - 10, warning + 10)
+        plt.show()
+
+
+plot_min_distances("Nov-12-2020_18-30-40")
```
tylerlum, line 40: Change this to be an argument. Look up how to make Python script with argument.

Can do this with simple python instead of ROS.

https://www.tutorialspoint.com/python/python_command_line_arguments.htm

tylerlum (approved):
I approve, after you change the hardcoded input folder name. 

Also, please add instructions to run this in the README.

Issue #144, opened 2020-10-16
Develop PID (or P) controller for heading and speed


tylerlum:
Our speed and heading are always exactly what is desired, which is not realistic, especially since we project obstacles based on our speed now....

We should model changes in heading and speed. Heading should not change instantly. Speed should change based on relative wind angle and wind speed. Speed should drop when we turn.

PR #162, opened 2020-11-27
Add smooth changes in heading and speed

haakov:
Fixes https://github.com/UBCSailbot/local-pathfinding/issues/144, replaces https://github.com/UBCSailbot/local-pathfinding/pull/155 (draft).

This PR introduces smooth changes in heading and speed using two P(I)-style controllers. The speed is reduced when turning more than 45 degrees. The maximum reduction factor is 3, for turns larger than or equal to 135 degrees.

Side effect: The speed will nearly drop to zero (but rise again) when the boat is in an invalid start state (with a trailing ship, for example) due to the planner sending lots of desiredHeading messages.

### DIFF
--- a/python/MOCK_controller_and_boat.py
+++ b/python/MOCK_controller_and_boat.py
@@ -10,8 +10,18 @@ from utilities import headingToBearingDegrees, PORT_RENFREW_LATLON
 
 # Constants
 GPS_PUBLISH_PERIOD_SECONDS = 0.1  # Keep below 1.0 for smoother boat motion
+BOAT_SPEED_KMPH = 14.4  # Boat should move at about 4m/s = 14.4 km/h
+TURNING_GAIN = 0.25  # Proportional term in the heading controller
+SPEED_CHANGE_GAIN = 0.05  # The same, but for the speed
+
 # Bring this value to about 1.0 for limited deviation. 20.0 for quite large deviations.
 HEADING_DEGREES_STANDARD_DEVIATION = 1.0
+
+# TODO: Move this to utilities
+# Smallest signed angle
+def ssa(angle):
+    return ((angle + 180) % 360) - 180
+
 
 
 class MOCK_ControllerAndSailbot:
@@ -20,7 +30,9 @@ class MOCK_ControllerAndSailbot:
         self.lon = lon
         self.speedup = 1.0
         self.headingDegrees = headingDegrees
+        self.headingSetpoint = headingDegrees
         self.speedKmph = speedKmph
+        self.speedSetpoint = speedKmph
         self.publishPeriodSeconds = GPS_PUBLISH_PERIOD_SECONDS
 
         rospy.init_node('MOCK_Sailbot_Listener', anonymous=True)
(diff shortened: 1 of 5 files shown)


python/MOCK_controller_and_boat.py
```diff
@@ -10,17 +10,29 @@
 
 # Constants
 GPS_PUBLISH_PERIOD_SECONDS = 0.1  # Keep below 1.0 for smoother boat motion
+BOAT_SPEED_KMPH = 14.4  # Boat should move at about 4m/s = 14.4 km/h
```
tylerlum, line 13: Great clear variable names

python/MOCK_controller_and_boat.py
```diff
@@ -10,17 +10,29 @@
...
+TURNING_GAIN = 0.25  # Proportional term in the heading controller
+SPEED_CHANGE_GAIN = 0.05  # The same, but for the speed
+
 # Bring this value to about 1.0 for limited deviation. 20.0 for quite large deviations.
 HEADING_DEGREES_STANDARD_DEVIATION = 1.0
 
+# TODO: Move this to utilities
```
tylerlum, line 20: Good idea to move this to utilities :)

tylerlum: If moved to utilities, might be good to specify degrees in func name or var name

python/MOCK_controller_and_boat.py
```diff
@@ -39,9 +51,23 @@ def move(self):
...
+        self.speedKmph += gain * error
+
     def desiredHeadingCallback(self, data):
         rospy.loginfo(data)
-        self.headingDegrees = data.headingDegrees + random.gauss(0, HEADING_DEGREES_STANDARD_DEVIATION)
+        self.headingSetpoint = data.headingDegrees
+        error_mag = abs(ssa(self.headingSetpoint - self.headingDegrees))
+        # Anything >= 135 has maximum effect (speed is divided by 3)
+        # Anything <= 45 has no effect on speed
+        self.speedKmph = self.speedKmph / max(1, min(3, error_mag / 45))
```
tylerlum, line 70: Potentially change speed setpoint based on wind direction

python/MOCK_controller_and_boat.py
```diff
@@ -39,9 +51,23 @@ def move(self):
         self.lon = destination.longitude
         self.lat = destination.latitude
 
+        # Heading
+        error = ssa(self.headingSetpoint - self.headingDegrees)
```
tylerlum, line 55: Great work, looks clear to me

tylerlum (changes requested):
Looks great! Added a couple of comments. 

Also, it would be great if this could be turned on/off with a rostopic or rosparam (might be easier with rosparam)

python/MOCK_controller_and_boat.py
haakov: Done! :)

haakov:
Wind setpoint is now divided by four when going upwind or downwind

tylerlum (approved):
Sorry totally forgot to review this again. Looks great! Tried it out and seems to work nicely. Just need to fix merge conflicts then good to go!

tylerlum:
And fix lint :)

PR #164, opened 2020-11-27
Fix speedup bug, so it doesn't wait if no speedup value is set

tylerlum:
This ensures that if we run `rosrun local_pathfinding main_loop.py`, then it won't wait for MOCK nodes to be setup for speedup publishing.


haakov (approved):
Looks good to me!

Issue #150, opened 2020-10-29
Implement and test interface node


tylerlum:
Implement a ROS node that takes in `Sensor.msg`s and outputs the topics that our ROS nodes use. This Sensor.msg is the information we will be receiving from the actual sensors. This interface node allows us to keep the same code we have developed

Link to the sensor message. https://github.com/UBCSailbot/sailbot-msg/blob/09056b6a496754dcb8653277063a604b5ae800c0/msg/Sensors.msg

Also, write a MOCK_allSensors.py script to simulate those inputs to see if everything is working as expected

tylerlum:
ROS NODE

Subscribe to topic of type Sensor.msg

Process/parse the information in the Sensor.msg, then publish to:
* GPS
* windSensor
* etc.

tylerlum:
The Sensor.msg topic name will be "sensors"

tylerlum:
Communicate with me and Alex Macdonald if you have questions

tylerlum:
https://github.com/UBCSailbot/local-pathfinding/blob/master/python/utilities.py

Take a look at the degrees to ensure we are in the right "frames" (what 0 degrees is and what +ve direction means)

tylerlum:
Completed by Patrick :)

PR #165, opened 2020-11-28
Ros Interface Node

patrick-5546:
Addresses issue #150.


python/ros_interace.py
```diff
@@ -0,0 +1,78 @@
...
+        self.pubWind = rospy.Publisher('windSensor', windSensor, queue_size=4)
+        self.pubGPS = rospy.Publisher('GPS', GPS, queue_size=4)
+
+        # sensorMsg = rospy.wait_for_message('sensors', Sensors)
+        self.wind_speedKmph = 0
+        self.wind_direction = 0
+        self.gps_lat = 0
+        self.gps_lon = 0
+        self.gps_headingDegrees = None
+        self.gps_speedKmph = 0
```
patrick-5546, line 29: Not sure what I should initialise these to.

python/ros_interace.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def checkErrAndAvg(self, arr):
+        # err_free_arr = checkError(arr)
+        err_free_arr = arr
+        return float(sum(err_free_arr)) / len(err_free_arr)
+
+    # def checkError(self, arr):
+    #     # input is an array of values
+    #     # returns array with values outside error of margin removed
+    #     # possible to modify a boolean if error is found?
```
patrick-5546, line 70: Is it necessary to check for errors, and not include sensor values that are outside a margin of error? If it is, I am not sure how to implement this behaviour.

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+    def checkError(self, arr):
+        pastValue = arr[0]
+        presentValues = arr - arr[0]
+
+        errorFreeList = list()
+        for value in presentValues:
+            if abs(float((pastValue - value)) / pastValue) < pastValue * ERROR:
+                errorFreeList.append(value)
+
+        return numpy.array(errorFreeList)
```
patrick-5546, line 99: Is it necessary to check for errors, and not include sensor values that are outside a margin of error? Seems like it could cause some errors, which I mentioned in line 88

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
+import rospy
+import numpy
+from sailbot_msg.msg import Sensors, windSensor, GPS
+
+# Constants
+CHECK_PERIOD_SECONDS = 0.1  # How often fields are updated
+KNOTS_TO_KMPH = 1.852
+ERROR = 0.2  # If signals need different margin of errors, modify as required
+
+# Conversion Notes:
```
tylerlum, line 10: Great notes! Very helpful. Can replace
```
# Comment
# Comment
```

with 
```
"""
Comment
Comment
```

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class RosInterface:
+    def __init__(self):
+        rospy.init_node('ros_interface', anonymous=True)
+        rospy.Subscriber("sensors", Sensors, self.sensorsCallback)
+        self.pubWind = rospy.Publisher('windSensor', windSensor, queue_size=4)
+        self.pubGPS = rospy.Publisher('GPS', GPS, queue_size=4)
+
+        # sensorMsg = rospy.wait_for_message('sensors', Sensors)
```
tylerlum, line 24: Remove this

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+
+        self.past_wind_speedKmph = 0
+        self.past_wind_direction = 0
+        self.past_gps_lat = 0
+        self.past_gps_lon = 0
+        self.past_gps_headingDegrees = 0
+        self.past_gps_speedKmph = 0
+
+    def sensorsCallback(self, data):
+        self.data = data
```
tylerlum, line 40: Set flag here

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+        self.gps_lon = 0
+        self.gps_headingDegrees = 0
+        self.gps_speedKmph = 0
+
+        self.past_wind_speedKmph = 0
+        self.past_wind_direction = 0
+        self.past_gps_lat = 0
+        self.past_gps_lon = 0
+        self.past_gps_headingDegrees = 0
+        self.past_gps_speedKmph = 0
```
tylerlum, line 37: Initialize flag as false

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def sensorsCallback(self, data):
+        self.data = data
+
+    def pub(self):
+        self.translate()
```
tylerlum, line 43: Check if flag is set.
If so, then continue
Else, ros warn about not having received first sensor message yet.

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+        # Past values
+        self.past_gps_lat = self.gps_lat
+        self.past_gps_lon = self.gps_lon
+        self.past_gps_headingDegrees = self.gps_headingDegrees
+        self.past_gps_speedKmph = self.gps_speedKmph
+        self.past_wind_speedKmph = self.wind_speedKmph
+        self.past_wind_direction = self.wind_direction
+
+        # GPS sensors
+        self.gps_lat = self.checkErrAndAvg([self.past_gps_lat,
```
tylerlum, line 58: Great clean code here. Very clear what you are doing

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+        self.wind_speedKmph = self.checkErrAndAvg([self.past_wind_speedKmph,
+                                                   self.data.wind_sensor_0_speed * KNOTS_TO_KMPH,
+                                                   self.data.wind_sensor_1_speed * KNOTS_TO_KMPH,
+                                                   self.data.wind_sensor_2_speed * KNOTS_TO_KMPH])
+        self.wind_direction = self.checkErrAndAvg([self.past_wind_direction,
+                                                   self.data.wind_sensor_0_direction,
+                                                   self.data.wind_sensor_1_direction,
+                                                   self.data.wind_sensor_2_direction])
+
+    # (0 is north, 90 is east, etc - cw - degrees) -> (0 is east, 90 is north, etc - ccw - degrees)
```
tylerlum, line 78: Great work, super clear

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+    def convertDegrees(self, degree):
+        converted = -1 * degree + 90
+        return converted if converted >= 0 else converted + 360
+
+    def checkErrAndAvg(self, arr):
```
tylerlum, line 83: Avoid the division by 0. For now, can do a check:

if len == 0:
WARN: Non matching values for GPS, defaulting to previous (or GPS 0)
Then return the first one or something.

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+        err_free_arr = self.checkError(arr)
+        return float(sum(err_free_arr)) / len(err_free_arr)
+
+    # past value is first element of arr
+    # might causes errors: returns empty array, what if sensor readings aren't consistent
+    # could modify past value to be the average of the past few outputs (store in array)
+    def checkError(self, arr):
```
tylerlum, line 90: I would change this name, as checkError isn't clear. Maybe call it something like getTrustedValues or getLowErrorValues. checkError makes me think it will return a bool

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+        pastValue = arr[0]
```
tylerlum, line 91: Seems a bit sketchy to just put pastValue as [0], maybe put it as a different parameter

def getTrustedValues(new_values, previous_value):

python/ros_interface.py
```diff
@@ -0,0 +1,107 @@
...
+        presentValues = arr - arr[0]
+
+        errorFreeList = list()
+        for value in presentValues:
+            if abs(float((pastValue - value)) / pastValue) < pastValue * ERROR:
```
tylerlum, line 96: Ensure that pastValue is non zero

python/ros_interface.py
tylerlum: Yeah I think what you have here is good for now, but it will definitely need to change later.

tylerlum (changes requested):
Great work! Have a couple of suggestions.

python/ros_interface.py
```diff
@@ -0,0 +1,114 @@
...
+    def getTrusted(self, pastValue, vals):
+        errorFreeVals = []
+        if pastValue != 0:
+            for value in vals:
+                if abs(float((pastValue - value)) / pastValue) < pastValue * ERROR:
+                    errorFreeVals.append(value)
+        else:
+            errorFreeVals = vals
+
+        return errorFreeVals
```
patrick-5546, line 106: Do you think this behaviour is fine for now? The assumptions I made are listed in line 95.

python/ros_interface.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def pub(self):
+        if self.initialized:
+            self.translate()
+            self.pubGPS.publish(self.gps_lat, self.gps_lon, self.gps_headingDegrees, self.gps_speedKmph)
+            self.pubWind.publish(self.wind_direction, self.wind_speedKmph)
+
+    # Translate from many sensor outputs to one signal, checking error margin, averaging, and converting units
+    def translate(self):
+        # Past values
```
tylerlum, line 53: One other thing I noticed. The sensorsCallback can happen at any moment, so we wouldn't want the self. data to change in the middle of this function, so you should store it in a local variable like 

`data = self.data` here, then use `data` instead of `self.data` throughout

python/ros_interface.py
```diff
@@ -0,0 +1,110 @@
...
+# Constants
+CHECK_PERIOD_SECONDS = 0.1  # How often fields are updated
+KNOTS_TO_KMPH = 1.852
+ERROR = 0.2  # If signals need different margin of errors, modify as required
+
+"""
+Conversion Notes:
+- wind_speedKmph, wind_direction, wind_headingDegrees, gps_speedKmph: int32 -> float64 (no loss of precision)
+- wind_speedKmph, gps_speedKmph: knots -> km/h
+- wind_direction: no conversion required (0 is bow, 90 is starboard/right, etc - cw - degrees)
```
tylerlum, line 13: I think this actually isn't right:

From utilities.py:
```
# Constants for boat frame angles
BOAT_RIGHT = 0
BOAT_FORWARD = 90
BOAT_LEFT = 180
BOAT_BACKWARD = 270
```

0 is right, not 90 in our code. I also think that our code treats it as ccw. Need to change this comment and add another conversion to the code.

python/ros_interface.py
```diff
@@ -0,0 +1,110 @@
...
+    def sensorsCallback(self, data):
+        if not self.initialized:
+            self.initialized = True
+        self.data = data
+
+    def pub(self):
+        if self.initialized:
+            self.translate()
+            self.pubGPS.publish(self.gps_lat, self.gps_lon, self.gps_headingDegrees, self.gps_speedKmph)
+            self.pubWind.publish(self.wind_direction, self.wind_speedKmph)
```
tylerlum, line 49: Can you add a warn if it is not initialized?

something like
````
else:
warn("Tried to publish sensor values, but not initialized yet. Waiting for first sensor message")
````

File: python/ros_interface.py [hunk 19]
```diff
@@ -0,0 +1,110 @@
...
+            return vals[0]
+        else:
+            return float(sum(err_free_vals)) / len(err_free_vals)
+
+    # assumes sensor readings are consistent, first sensor initial reading is accurate
+    # could modify past value to be the average of the past few outputs (store in list)
+    def getTrusted(self, pastValue, vals):
+        errorFreeVals = []
+        for value in vals:
+            if abs(float((pastValue - value)) / pastValue) < pastValue * ERROR:
```
tylerlum, line 100: I think you left in the `/ pastValue` part by accident. I see what you were doing to avoid division by 0 and still comparing the same, but it's not quite as clear. Maybe like

```
if pastValue != 0 and abs(float(pastValue - value) / pastValue) < ERROR:
```

tylerlum (approved):
Looks good to me. Address my comments and then good to merge!

python/ros_interface.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class RosInterface:
+    def __init__(self):
+        rospy.init_node('ros_interface', anonymous=True)
+        rospy.Subscriber("sensors", Sensors, self.sensorsCallback)
+        self.pubWind = rospy.Publisher('windSensor', windSensor, queue_size=4)
+        self.pubGPS = rospy.Publisher('GPS', GPS, queue_size=4)
+        self.initialized = False
+        self.data = 0
```
patrick-5546, line 25: I forgot to initialise self.data before

python/ros_interface.py
```diff
@@ -0,0 +1,115 @@
...
+            self.pubWind.publish(self.wind_direction, self.wind_speedKmph)
+        else:
+            rospy.logwarn("Tried to publish sensor values, but not initialized yet. Waiting for first sensor message.")
+
+    # Translate from many sensor outputs to one signal, checking error margin, averaging, and converting units
+    def translate(self):
+        data = self.data
+
+        # Past values
+        self.past_gps_lat = self.gps_lat
```
tylerlum, line 59: Sorry one more thing. On first loop, these will all be 0's, which doesn't make sense. The error checking code will be messed up. You already have a check for the value to be 0, but doesn't work for the conversions (turns 0 to 90). 

Could init them all to `None` instead of 0, then have a None check in Err function

python/ros_interface.py
```diff
@@ -0,0 +1,115 @@
...
+            return vals[0]
+        else:
+            return float(sum(err_free_vals)) / len(err_free_vals)
+
+    # assumes sensor readings are consistent, first sensor initial reading is accurate
+    # could modify past value to be the average of the past few outputs (store in list)
+    def getTrusted(self, pastValue, vals):
+        errorFreeVals = []
+        for value in vals:
+            if pastValue != 0 and abs(float((pastValue - value)) / pastValue) < ERROR:
```
tylerlum, line 105: `if pastValue is not None and...`

tylerlum (approved):

Issue #171, opened 2020-12-11
Change both local and global waypoint reached to use the line-based method


tylerlum:
Task 1: Change Path waypointReached so that there is some distance between the waypoint and the line. (hard to word, look at examples below)

https://github.com/UBCSailbot/local-pathfinding/blob/master/python/Path.py#L341

WHAT WE WANT:
```
        Examples: B is boat. 0 is (self._nextWaypointIndex - 1). 1 is (self._nextWaypointIndex).

        Example 1 Waypoint not reached:
              |
              |
        0  B    1
              |
              |

        Example 2 Waypoint reached:
              |
              |
        0       1
              |
              |B
```

WHAT IT IS NOW:
```
        Examples: B is boat. 0 is (self._nextWaypointIndex - 1). 1 is (self._nextWaypointIndex).

        Example 1 Waypoint not reached:
                |
                |
        0  B    1
                |
                |

        Example 2 Waypoint reached:
                |
                |
        0       1
                |
                |B
```

Task 2: Replace globalWaypointReached (radius based) to be the same offset line-based approach as local waypoint reached.

https://github.com/UBCSailbot/local-pathfinding/blob/master/python/utilities.py#L130
^Likely remove this entirely

https://github.com/UBCSailbot/local-pathfinding/blob/master/python/main_loop.py#L115
^ likely replace

```
        isGlobalWaypointReached = utils.globalWaypointReached(state.position, state.globalWaypoint)
```

with 
```
        isGlobalWaypointReached = localPath.lastWaypointReached(state.position)
```

tylerlum:
![image](https://user-images.githubusercontent.com/26510814/101868996-4f9d1e80-3b33-11eb-8642-d8571c30508d.png)

tylerlum:
Resolved in #172

PR #172, opened 2020-12-11
Revise waypoint reached

patrick-5546:
Addresses #171.


python/Path.py
```diff
@@ -430,6 +438,19 @@ def x(y):
...
+        nextWaypointLatlon = self._latlons[self._nextWaypointIndex]
+
+        return self.waypointReached(positionLatlon, previousWaypointLatlon, nextWaypointLatlon)
+
+    # assumes there are at least 2 elements in self._latlons
+    def lastWaypointReached(self, positionLatlon):
+        previousWaypointLatlon = self._latlons[len(self._latlons) - 2]
+        nextWaypointLatlon = self._latlons[len(self._latlons) - 1]
+
+        return self.waypointReached(positionLatlon, previousWaypointLatlon, nextWaypointLatlon)
```
patrick-5546, line 452: Issues with this implementation:
* Returns true every time the simulation starts
* Assumes that there are at least 2 elements in `self._latlons`

tylerlum (commented):
Looks good to me so far. Update the unit tests to ensure proper functionality. I will do some testing myself and approve after :)

tylerlum (changes requested):
Great work. Make sure to add in those unit tests to gain confidence in the code, and double check the verticalShift math.

tylerlum (approved):
Works beautifully. Might still be a good idea to draw the boundary line in another PR. Go ahead and merge.

PR #175, opened 2021-01-15
In the event of no recent wind sensor messages, set the wind to 90 deg

haakov:
It looks like the local/global wind calculation is incorrect, I'll investigate this

### DIFF
--- a/python/main_loop.py
+++ b/python/main_loop.py
@@ -22,6 +22,12 @@ localPathUpdateForced = False
 # Global variable for speedup
 speedup = 1.0
 
+# Global variable for when the last wind sensor message arrived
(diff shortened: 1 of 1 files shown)


tylerlum:
Let me know which part of the local/global wind calculation may be incorrect. I thought that too a few months ago, but after some debugging, I concluded it's actually fine. But could you describe the issue with some screenshots and/or diagrams so we can discuss?

haakov:
PR should be good now, continuing discussion in https://github.com/UBCSailbot/local-pathfinding/issues/176

tylerlum:
Closing this in place of #183

Issue #163, opened 2020-11-27
Add line of sight (LOS) guidance law


haakov:
Adding a LOS guidance law will keep the boat from drifting off track, but it will also mean that we lose control (slightly) of the desired heading that the local planner is passing on to the controller.

<img width="576" alt="los_guidance" src="https://user-images.githubusercontent.com/1413378/100416197-21710680-307e-11eb-9a08-5b973e0bcacf.png">
Variables:

-  y_p^e - cross-track error
- alpha - track heading
- theta - heading output (input to the controller)
- phi - lookahead distance

haakov:
To illustrate, here is a picture of our current pathfinding algorithm after 30 seconds with speedup=10 and current=5 kmph eastward:
<img src="https://user-images.githubusercontent.com/1413378/106128766-f5e2c900-615f-11eb-8b9d-dfcd13ab8413.png" width="500" height="500" />
As we can see, the boat has drifted slightly off track and is currently heading towards the next waypoint. Here is a screenshot with LOS guidance enabled:
<img src="https://user-images.githubusercontent.com/1413378/106128749-f11e1500-615f-11eb-9cff-b1d30de27b38.png" width="500" height="500" />
The LOS guidance law adds a small (tuneable) angle offset to the desired heading in order to keep the boat on track.

PR #178, opened 2021-01-28
Add optional LOS guidance

haakov:
More info: https://github.com/UBCSailbot/local-pathfinding/issues/163


tylerlum:
image

Tried to push/pull and merge but for some reason the diff isn't showing up right.

image

Starting new PR shows the right diff though ...

PR #183, opened 2021-02-19
Missing and stale data

tylerlum:
The purpose of this PR is to handle missing or stale sensor/global path data.

**Problem 1**: The pathfinding should not start until Sailbot.py has received its first GPS, AIS, windSensor, and globalPath message (otherwise, it would be using defaults that are completely arbitrary). 

**Solution 1**: Before starting main_loop, wait to receive first messages. Start all values as None to be very clear.

**Problem 2**: The pathfinding should send obvious warnings when it does not receive data for a certain time. Eg. if it doesn't receive windSensor data for 60 seconds, then we should know that something is wrong.

**Solution 2**: Set a `lastUpdated` member variable for each sensor/global path callback. Each time we getCurrentState of Sailobt, we should get warnings if there is stale data.

To test, don't use roslaunch, but open one node at a time.

```
roscore
```

```
rosrun local_pathfinding main_loop.py
```

Should now display that it is waiting for data.

```
rosrun local_pathfinding MOCK_controller_and_boat.py
```

Display should show that GPS data has been received.

```
rosrun local_pathfinding MOCK_wind_sensor.py
```

```
rosrun local_pathfinding MOCK_global_planner.py
```

```
rosrun local_pathfinding MOCK_AIS.py
```

Now pathfinding should begin.

Run the visualizer to view what's going on:

```
rosrun local_pathfinding local_path_visualizer.py
```

Then close one of the MOCK nodes to simulate a sensor no longer sending data. Should get yellow warning messages after the time limit has been reached (60s right now).

### DIFF
--- a/python/Sailbot.py
+++ b/python/Sailbot.py
@@ -7,6 +7,19 @@ import sys
 
 
 class BoatState:
+    """Datatype representing the current state of the sailbot, which includes all information relevant for pathfinding.
+
+    Attributes:
+        position (sailbot_msg.msg._latlon.latlon): latlon that is the current latlon position of the boat
+        measuredWindDirectionDegrees (float): measured direction that the wind is going towards wrt the boat
+                                              (0 degrees = going to right of boat, 90 degrees = going to front of boat)
+        measuredWindSpeedKmph (float): measured speed of the wind wrt to boat
+        AISData (sailbot_msg.msg._AISMsg.AISMsg): AIS message with information about nearby boats
+        headingDegrees (float): direction that the boat is heading towards
+                                (0 degrees = going east, 90 degrees = going north)
+        speedKmph (float): speed that the boat is moving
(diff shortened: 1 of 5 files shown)


tylerlum:
I did a review of @haakov 's #175 PR, which I think was good in principle. However, I thought it made sense to:

1. Make this behavior apply to all sensor/global path (AIS, GPS, windSensor, globalPath)
2. Keep main_loop.py clean by abstracting this all in Sailbot.py

haakov:
Makes a lot of sense! So to be clear, this PR replaces https://github.com/UBCSailbot/local-pathfinding/pull/175, right?

tylerlum:
Yes that's the idea. Need to fix the unit tests tho

tylerlum:
Ready for review

patrick-5546 (approved):

PR #184, opened 2021-02-19
Fixed Global Waypoint Reached Line

### LINKED ISSUE #180: Fix global waypoint reached normal line
Right now, the global waypoint reached function checks if the boat has cross the normal line between the second last local waypoint and the last local waypoint. However, this is not robust to big changes in GPS, as the path would then change. The proper solution is the set this normal line between the PREVIOUS GLOBAL WAYPOINT and the CURRENT GLOBAL WAYPOINT.

patrick-5546:
Resolves #180. In short, the global waypoint reached line (red) was made distinct from the local waypoint reached lines (blue), as it should be with respect to the global waypoints instead of the local waypoints.

### DIFF
--- a/python/Path.py
+++ b/python/Path.py
@@ -344,6 +344,28 @@ class Path:
                 return path[pathIndex]
 
         return _getNextWaypoint(self._latlons, self._nextWaypointIndex)
+    
+    def getPreviousWaypoint(self):
+        '''Gets the previous waypoint, but performs bounds and edge case checks'''
+        def _getPreviousWaypoint(path, pathIndex):
+            # If path is empty, return (0, 0)
+            if len(path) == 0:
+                rospy.logwarn("Path is empty.")
+                rospy.logwarn("Setting waypoint to be (0, 0).")
+                return latlon(0, 0)
+
+            # If index out of range, return last waypoint in path
+            if pathIndex >= len(path):
+                rospy.logwarn("Path index is out of range: index = {} len(path) = {}".format(pathIndex, len(path)))
+                rospy.logwarn("Setting waypoint to be the last element of the path")
+                pathIndex = len(path) - 1
+                return path[pathIndex]
+
+            # If index in range, return the correct waypoint
+            else:
+                return path[pathIndex]
+
+        return _getPreviousWaypoint(self._latlons, self._nextWaypointIndex - 1)
 
     # Simple methods that directly call _omplPath methods
     def getCost(self):
--- a/python/Path.py
+++ b/python/Path.py
@@ -460,6 +460,13 @@ class Path:
 
         return self.waypointReached(positionLatlon, previousWaypointLatlon, nextWaypointLatlon)
 
+    # assumes there are at least 2 elements in self._latlons
+    def lastWaypointReached(self, positionLatlon):
+        previousWaypointLatlon = self._latlons[-2]
+        nextWaypointLatlon = self._latlons[-1]
+
+        return self.waypointReached(positionLatlon, previousWaypointLatlon, nextWaypointLatlon)
+
     def _cleanNumLookAheadWaypoints(self, numLookAheadWaypoints):
         '''Ensure nextLocalWaypointIndex + numLookAheadWaypoints is in bounds
         Default behavior when numLookAheadWaypoints is not given OR bad input: set to max
--- a/python/Path.py
+++ b/python/Path.py
@@ -344,7 +344,7 @@ class Path:
                 return path[pathIndex]
 
         return _getNextWaypoint(self._latlons, self._nextWaypointIndex)
-    
+
     def getPreviousWaypoint(self):
         '''Gets the previous waypoint, but performs bounds and edge case checks'''
         def _getPreviousWaypoint(path, pathIndex):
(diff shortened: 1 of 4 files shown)


tylerlum (changes requested):
Clean changes and seems to work wonderfully. Would be great if you could add a line to the visualizer.

tylerlum:
How to test:

```
roslaunch local_pathfinding all.launch random_seed:=100
```

Then move the boat way past the global waypoint. For small distance, (use tabs, not copy and paste what I have here)

```
rostopic pub /changeGPS sailbot_msg/GPS "lat: 48
lon: -125
headingDegrees: 0.0
speedKmph: 14.0"
```

For longer distance

```
rostopic pub /changeGPS sailbot_msg/GPS "lat: 46
lon: -126
headingDegrees: 0.0
speedKmph: 14.0"
```

Before, would trek backwards. Now when it notices it would be going backwards, it updates the next global waypoint. Does this over and over again until it starts moving forwards. Would be great to see a visual of this line.

tylerlum:
image
image
image

When teleports forward, decides to go forwards to next waypoint

patrick-5546 (commented):
Red is global waypoint reached line, blue is local waypoint reached line. 
Screenshot from 2021-02-19 18-21-13
One thing to note is that there is a blue and red line graphed as the boat is reaching the global waypoint. Should the blue line not be graphed for the last waypoint?
Screenshot from 2021-02-19 18-28-56

patrick-5546:
Might need some additional error checking for the new `previousGlobalWaypoint` and `previousLocalWaypoint` ROS nodes

tylerlum:
Great work. I see what you mean with the error checking for previous waypoints. However, we always set the globalPathIndex and localPathIndex to 1 at start, and it only either increases or restarts at 1. Could still write some checks to be safe, but I think it is pretty okay already

python/Path.py
```diff
@@ -29,35 +29,35 @@
 temp_invalid_solutions = 0
 
 
-def getPerpLine(lastX, lastY, nextX, nextY):
+def getPerpLine(lastXY, nextXY):
```
tylerlum, line 32: I like that you made the input args cleaner. Bit of a nitpick, but I find it easier to understand
````
if abs(nextY - lastY) < 0.01:
```

More easily than

```
if abs(nextXY[1] - lastXY[1]) < 0.01:
```

Consider keeping the same args, but then in the first line of the function, make `nextX, nextY = nextXY`, and then using nextX instead of nextXY[0]

tylerlum (approved):
Just teste it out, works great! One small comment i added, not needed if you don't agree :)

patrick-5546:
Something to consider: what happens when the next local waypoint is the next global waypoint, and the local waypoint is reached but the global waypoint isn't? In the screenshot below, when the boat is past the blue line but before the red line.
Screenshot from 2021-02-20 14-14-00

tylerlum:
Investigated the between red and blue line issue. 

* If crosses red line first, no problem. Goes to next global waypoint
* If crosses blue line first, then will keep generating a new path every loop, leading to a lot of wasteful energy usage and slow simulation.

Proposed solution: if EITHER crosses red OR crosses the FINAL blue, then go to next global waypoint

python/main_loop.py
```diff
@@ -112,7 +114,9 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
             state, numLookAheadWaypoints=utils.NUM_LOOK_AHEAD_WAYPOINTS_FOR_UPWIND_DOWNWIND, showWarnings=True)
         hasObstacleOnPath = localPath.obstacleOnPath(
             state, numLookAheadWaypoints=utils.NUM_LOOK_AHEAD_WAYPOINTS_FOR_OBSTACLES, showWarnings=True)
-        isGlobalWaypointReached = localPath.lastWaypointReached(state.position)
+        nextGlobalWaypoint = sailbot.globalPath[sailbot.globalPathIndex]
+        previousGlobalWaypoint = sailbot.globalPath[sailbot.globalPathIndex - 1]
+        isGlobalWaypointReached = localPath.waypointReached(state.position, previousGlobalWaypoint, nextGlobalWaypoint)
```
tylerlum, line 119: Could update this as:

```
            isLocalWaypointReached = localPath.nextWaypointReached(state.position)
            isClosestToEnd = isClosestToLastWaypoint(localPath) (maybe simple for loop over latlons?)
            isGlobalWaypointReached = localPath.waypointReached(state.position, previousGlobalWaypoint, nextGlobalWaypoint) or (isLocalWaypointReached and isClosestToEnd)
```

tylerlum, line 119: Would also be great to have a different hardcoded value for global and local waypoint reached, so we can more easily test and debug. Maybe add a default parameter of distance or a default parameter bool of true or false?

tylerlum (changes requested):

tylerlum:
Probably best to work on this AFTER merging in #186

patrick-5546:
Both `GLOBAL_WAYPOINT_REACHED_DISTANCE` (red line) and `WAYPOINT_REACHED_DISTANCE` (blue lines) are currently set to 0.5, but can be easily altered to optimize pathfinding algorithm.

tylerlum (approved):
Tested this out just now, works beautifully.

Set
```
WAYPOINT_REACHED_DISTANCE = 1.5
GLOBAL_WAYPOINT_REACHED_DISTANCE = 0.5
```

Was able to ensure that passing last blue line meant globalWaypointReached.

Then used rostopic pub /changeGPS to move the boat and ensure that the red line was working as well.

PR #185, opened 2021-02-19
Add optional LOS guidance (written by Hakon, but reopened)

tylerlum:
More info: #163

### DIFF
--- a/python/main_loop.py
+++ b/python/main_loop.py
@@ -68,6 +68,9 @@ if __name__ == '__main__':
     # Create sailbot ROS object that subscribes to relevant topics
     sailbot = sbot.Sailbot(nodeName='local_pathfinding')
 
+    # Get enable_los parameter
+    enable_los = rospy.get_param('enable_los', default=False)
+
     # Subscribe to requestLocalPathUpdate
     rospy.Subscriber('requestLocalPathUpdate', Bool, localPathUpdateRequestedCallback)
     rospy.Subscriber('forceLocalPathUpdate', Bool, localPathUpdateForcedCallback)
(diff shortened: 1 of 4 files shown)


python/utilities.py
```diff
@@ -162,6 +162,40 @@ def getDesiredHeadingDegrees(position, localWaypoint):
...
+       currentWaypoint (sailbot_msg.msg._latlon.latlon): Current local waypoint that the boat is aiming towards
+
+    Returns:
+       float that is the heading (degrees) that the boat should be aiming towards to reach the local waypoint
+    '''
+    # The lookahead distance describes how "hard" the boat should try to reduce the crosstrack error
+    lookahead = 0.001
+    Kp = 1 / lookahead
+
+    x_start = previousWaypoint.lon
```
tylerlum, line 181: Code is clean and easy to understand. 

One thing to note is that the units of lon and lat are not the same length, so angle calculation may not be right. Could use latlonToXY to calculate in km space.

python/main_loop.py
```diff
@@ -68,6 +68,9 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
     # Create sailbot ROS object that subscribes to relevant topics
     sailbot = sbot.Sailbot(nodeName='local_pathfinding')
 
+    # Get enable_los parameter
+    enable_los = rospy.get_param('enable_los', default=False)
```
tylerlum, line 72: Nice clean way to add this in

tylerlum (commented):
Code looks good. Just one comment about the latlon units.

Also, I noticed it often started upwind/downwind detected panicking when doing this. Can you test this out and show us how it behaves?

haakov:
Not sure about the panicking, I haven't seen that yet. What did you do to produce those results?

This is a gif showing the most obvious drawback. Using LOS guidance means that the boat's yaw angle will deviate from the direction found by the planner, meaning that the boat can end up going up/downwind. That said, the LOS guidance is tuneable, and the situation in the gif is perhaps a little contrived, with a significantly higher current speed than wind speed.
los

Short gifs showing the behaviour below (speedup=10, current=5 kmph eastward, wind 2-3 kmph northward)

## Without LOS
no_los_short

## With LOS
los_short

tylerlum:
Great analysis! I played around with it more and I'm happy with it. Code is clear and will written, and I like the new behaviour of it getting back on track. Fix lint and should be good to merge!

python/utilities.py
haakov: ah, of course! Totally forgot about the lon/lat length

tylerlum:
Merging now :)

PR #186, opened 2021-02-19
Debug waypoint reached

tylerlum:
Investigating bug with waypoint reached line and removepastwaypoints

### DIFF
--- a/python/Path.py
+++ b/python/Path.py
@@ -153,9 +153,8 @@ class OMPLPath:
     def removePastWaypoints(self, state):
         """Removes waypoints that the boat has already passed.
 
-        Note: Best used when waypointReached() == True, not every loop, otherwise the
-              localWaypointReached tangent line could get messed up.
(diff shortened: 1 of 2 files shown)


tylerlum:
@patrick-5546  fixed the issue with the waypoint reached line, so it is always the correct distance away.

TODO: Change behavior to probably never prepend boat position (do we need to do this in the case with lengthAfter==1)?

tylerlum:
Fixed 2 issues:
1. Waypoint reached line in wrong position. Fixed by changing sin to cos
2. Boat still kind of goes "backwards" at times, so we improve the removePastWaypoints function. Only potential issue: with high speedup, boat will overshoot the waypoint, so this will make the boat skip a waypoint. Not a big deal, but good to note.

tylerlum:
Ready for review.

patrick-5546 (approved):
Efficiently resolves the documented bug, glad this is working as expected now.

PR #187, opened 2021-03-10
Create Collision Buffer

### LINKED ISSUE #143: Account for boat speed changes/uncertainty
Right now we have a hardcoded boat speed, but it will not always be this way. We may want to model changes in boat speed. This could change how we project obstacles, as we need to be conservative about how we model their movement

patrick-5546:
Fixes #143


tylerlum (approved):
Looks great. Could improve naming of SAILBOT_MAX_SPEED_BUFFER to be a bit more clear (eg. units)

Issue #182, opened 2021-02-18
Add current boat speed to local path visualizer


tylerlum:
The visualizer does not clearly show the boat speed. The heading is easy to infer from the triangle's direction, but the boat speed is hard to infer, especially at large distance scales and at varying speedups.

We should have MOCK_controller_and_boat.py (every update) OR main_loop.py (every loop, state has speed) publish something like "boatSpeed" float, and the local path visualizer subscribe and print it to screen somewhere. Especially helpful working with #144 smooth boat speed changes.

tylerlum:
@haakov could you add this in? Would be great to have this to debug the smooth speed changes :)

PR #188, opened 2021-03-11
Show boat speed in visualizer

haakov:
Fixes https://github.com/UBCSailbot/local-pathfinding/issues/182

Uses the GPS message

speed


tylerlum:
@haakov just tried it out, and the speed stuff worked well! I ran it and found that the speedkmph value initially got cut off by the other text, so I had to resize the plot first. image.

I tried out moving the speed to the sailbot latlon text, what do you think?
image

haakov:
I like moving it to the latlon text!

tylerlum (approved):

PR #191, opened 2021-03-20
Move Global Waypoint when obstacle in the way

### LINKED ISSUE #190: Debug and resolve issue when sailbot tries to go to very far away global waypoint
image

Sometimes the sailbot tries to go to a very far away waypoint. Should be going to waypoint ~30km away, but starts going 100km away or more. Potential reason:

From main_loop.py, we tell the sailbot to go to the next global waypoint if we reach the waypoint or if the global waypoint is invalid. Likely not from reaching next global waypoint, as this seems to be implemented properly. Issue is when there is a boat on the global waypoint, making it invalid. Then the system can't do pathfinding to that point, so we skip that waypoint. Likely find another solution like move the waypoint slightly left or right, or just ignore the boat unless 1 waypoint away (or n waypoints away).
image

patrick-5546:
Resolves #190 

See **Progress** comment for more details

### DIFF
--- a/python/Path.py
+++ b/python/Path.py
@@ -465,10 +465,11 @@ class Path:
 
     # assumes there are at least 2 elements in self._latlons
     def lastWaypointReached(self, positionLatlon):
-        previousWaypointLatlon = self._latlons[-2]
-        nextWaypointLatlon = self._latlons[-1]
+        # last waypoint reached implies the next waypoint is the last waypoint in self._latlons
+        if self._nextWaypointIndex != len(self._latlons) - 1:
+            return False
 
-        return self.waypointReached(positionLatlon, previousWaypointLatlon, nextWaypointLatlon)
+        return self.nextWaypointReached(positionLatlon)
 
     def _cleanNumLookAheadWaypoints(self, numLookAheadWaypoints):
         '''Ensure nextLocalWaypointIndex + numLookAheadWaypoints is in bounds
--- a/python/Sailbot.py
+++ b/python/Sailbot.py
@@ -135,7 +135,13 @@ class Sailbot:
             if not goalWasInvalid:
                 rospy.logwarn("Goal state invalid, finding new global waypoint")
                 rospy.loginfo("Old waypoint: {}".format(state.globalWaypoint))
-                movedGlobalWaypoint = his.moveGlobalWaypointUntilValid(state)
+
+                # calculating waypoint to use in moveGlobalWaypointUntilValid
+                isLast = self.globalPathIndex == len(self.globalPath) - 1
+                otherWaypointIndex = self.globalPathIndex - 1 if isLast else self.globalPathIndex + 1
+                otherWaypoint = self.globalPath[otherWaypointIndex]
+
+                movedGlobalWaypoint = his.moveGlobalWaypointUntilValid(state, isLast, otherWaypoint)
                 rospy.loginfo("Moved waypoint: {}".format(movedGlobalWaypoint))
                 goalWasInvalid = True
 
--- a/python/Sailbot.py
+++ b/python/Sailbot.py
@@ -1,9 +1,14 @@
 #!/usr/bin/env python
 import rospy
+import handleInvalidState as his
 from sailbot_msg.msg import AISMsg, GPS, path, latlon, windSensor
 from geopy.distance import distance
 import time
 import sys
+
+# Global variables for moving waypoint
+goalWasInvalid = False
+movedGlobalWaypoint = None
 
 
 class BoatState:
@@ -121,8 +126,22 @@ class Sailbot:
             rospy.logwarn("Setting globalWaypoint to be the last element of the globalPath")
             self.globalPathIndex = len(self.globalPath) - 1
 
-        return BoatState(self.globalPath[self.globalPathIndex], self.position, self.measuredWindDirectionDegrees,
-                         self.measuredWindSpeedKmph, self.AISData, self.headingDegrees, self.speedKmph)
+        state = BoatState(self.globalPath[self.globalPathIndex], self.position, self.measuredWindDirectionDegrees,
+                          self.measuredWindSpeedKmph, self.AISData, self.headingDegrees, self.speedKmph)
+
+        global goalWasInvalid
+        global movedGlobalWaypoint
+        if goalWasInvalid or not his.checkGoalValidity(state):
+            if not goalWasInvalid:
+                rospy.logwarn("Goal state invalid, finding new global waypoint")
+                rospy.loginfo("Old waypoint: {}".format(state.globalWaypoint))
+                movedGlobalWaypoint = his.moveGlobalWaypointUntilValid(state)
+                rospy.loginfo("Moved waypoint: {}".format(movedGlobalWaypoint))
+                goalWasInvalid = True
+
+            state.globalWaypoint = movedGlobalWaypoint
+
+        return state
 
     def GPSCallback(self, data):
         self.position = latlon(data.lat, data.lon)
--- a/python/Sailbot.py
+++ b/python/Sailbot.py
@@ -131,20 +131,21 @@ class Sailbot:
 
         global goalWasInvalid
         global movedGlobalWaypoint
-        if goalWasInvalid or not his.checkGoalValidity(state):
-            if not goalWasInvalid:
-                rospy.logwarn("Goal state invalid, finding new global waypoint")
-                rospy.loginfo("Old waypoint: {}".format(state.globalWaypoint))
 
-                # calculating waypoint to use in moveGlobalWaypointUntilValid
-                isLast = self.globalPathIndex == len(self.globalPath) - 1
-                otherWaypointIndex = self.globalPathIndex - 1 if isLast else self.globalPathIndex + 1
-                otherWaypoint = self.globalPath[otherWaypointIndex]
+        if not goalWasInvalid and not his.checkGoalValidity(state):
+            rospy.logwarn("Goal state invalid, finding new global waypoint")
+            rospy.loginfo("Old waypoint: {}".format(state.globalWaypoint))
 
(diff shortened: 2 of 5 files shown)


patrick-5546:
Progress:
* `getNewGlobalWaypoint()` works as expected, called in every iteration through `main_loop` once the goal state is invalid
    * However, updating the path every time is very slow. I get these warning in every iteration:
Screenshot from 2021-03-31 15-18-47
* Visualizer not updated yet, global waypoint marker and waypoint reached lines remain in their initial positions

python/main_loop.py
```diff
@@ -148,13 +148,16 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
...
             # If globalWaypoint reached, increment the index
-            if isGlobalWaypointReached or goalInvalid:
+            if isGlobalWaypointReached:
                 if isGlobalWaypointReached:
                     rospy.loginfo("Global waypoint reached")
-                if goalInvalid:
-                    rospy.logwarn("Goal state invalid, skipping to the next global waypoint")
                 sailbot.globalPathIndex += 1
 
+            # If goal invalid, use a new global waypoint
```
tylerlum, line 156: Great comment, could add at the end "move global waypoint toward/away from the sailbot until it is valid" to be extra clear

python/handleInvalidState.py
```diff
@@ -83,7 +85,7 @@ def generateSafeHeadingDegrees(state, invalidStartObstacle):
     return globalWindDirectionDegrees
 
 
-def checkGoalValidity(state):
+def checkGoalValidity(state, goalLatlon=None):
```
tylerlum, line 88: I like this solution! Could you add some documentation to the docstring right under here. Just say what the goalLatlon is and how it is used, and what happens with goalLatlon is None

python/handleInvalidState.py
```diff
@@ -92,10 +94,36 @@ def checkGoalValidity(state):
...
     obstacles = obs.getObstacles(state, referenceLatlon)
     for obstacle in obstacles:
-        goalXY = utils.latlonToXY(state.globalWaypoint, referenceLatlon)
+        goalXY = utils.latlonToXY(goalLatlon, referenceLatlon)
         if not obstacle.isValid(goalXY):
             return False
     return True
+
+
+def getNewGlobalWaypoint(state):
```
tylerlum, line 109: The function implementation looks really good to me, but a few comments:
1. Consider changing name to "getValidGlobalWaypoint" or "getClosestValidGlobalWaypoint" or "moveGlobalWaypointUntilValid", to make it clear it is not getting the NEXT waypoint, but getting the closest latlon to the global waypoint that is valid
2. Docstring is good, but more detail would be nice. Keep what you have. But also add something like "moves the global waypoint toward/away from sailbot until valid. Tbh just reading the code quickly, I'm not 100% sure if it moves TOWARD or AWAY from the boat, so that would be really helpful. Otherwise the code is very clean and easy to understand.
3. In this code, MOVE_GOAL_WAYPOINT_STEP_SIZE is very clear. But at the top of file, MOVE_GOAL_WAYPOINT_STEP_SIZE is a bit unclear what unit this is (unitless or km etc.). Consider adding _KM at the end

tylerlum (approved):
Looks really on the right track. Just added some comments that are less focused on function and more focused on some documentation/naming. Great work!

tylerlum:
I looked into this. Looks like you implemented it exactly like we had in mind, but it has some issues. It remakes a path every loop. Need to find a new way to do it.

python/Path.py
```diff
@@ -465,10 +465,11 @@ def nextWaypointReached(self, positionLatlon):
 
     # assumes there are at least 2 elements in self._latlons
     def lastWaypointReached(self, positionLatlon):
-        previousWaypointLatlon = self._latlons[-2]
-        nextWaypointLatlon = self._latlons[-1]
+        # last waypoint reached implies the next waypoint is the last waypoint in self._latlons
```
tylerlum, line 468: Perfect!

python/handleInvalidState.py
```diff
@@ -7,6 +7,8 @@
 import obstacles as obs
 import utilities as utils
 
+MOVE_GOAL_WAYPOINT_STEP_SIZE_KM = 5
```
tylerlum, line 10: Great name

python/handleInvalidState.py
```diff
@@ -83,19 +85,49 @@ def generateSafeHeadingDegrees(state, invalidStartObstacle):
     return globalWindDirectionDegrees
 
 
-def checkGoalValidity(state):
-    '''Checks if the goal waypoint of the given state is valid (no obstacle there)
+def checkGoalValidity(state, goalLatlon=None):
+    '''Checks if the goal of the given state is valid (no obstacle there)
```
tylerlum, line 89: Great docstring

python/handleInvalidState.py
```diff
@@ -83,19 +85,49 @@ def generateSafeHeadingDegrees(state, invalidStartObstacle):
...
+
+
+def moveGlobalWaypointUntilValid(state):
+    '''Moves global waypoint towards sailbot until it is valid
+
+    Args:
+       state (BoatState): State of the sailbot and other boats
+
+    Returns:
+        latlon goalLatlon such that checkGoalValidity(state, newGoal) is true
```
tylerlum, line 117: Could you add something in the docstring about whether the new waypoint will be further TOWARD or AWAY from sailbot? Eg. will new point move on straight line toward sailbot or away from sailbot. 

Eg. "draws straight line between sailbot and current global waypoint, then moves the goalLatlon along this line TOWARD (or AWAY? tbh unsure) until it is valid.

python/main_loop.py
```diff
@@ -148,12 +154,19 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
...
-                    rospy.loginfo("Global waypoint reached")
-                if goalInvalid:
-                    rospy.logwarn("Goal state invalid, skipping to the next global waypoint")
+            if isGlobalWaypointReached:
+                rospy.loginfo("Global waypoint reached")
                 sailbot.globalPathIndex += 1
+                goalWasInvalid = False
+
+            # If goal invalid, move global waypoint towards sailbot until it is valid
+            if goalInvalid and not goalWasInvalid:
```
tylerlum, line 163: I like your solution, it seems to be working better. BUT we still have the issue that the local path visualizer is a bit confusing. 

I realized that the function moveGlobalWaypointUntilValid() only depends on state. this whole update thing depends on both state and also the global variable goalWasInvalid. Could you look into putting this functionality all into sailbot class?

Then both main_loop and local_path_visualizer would use the sailbot class the same way and have it work. Eg. update sailbot.getCurrentState to automatically do this obstacle check and move of global waypoint. May also need to have some sort of class variable to be the equivalent of "goalWasInvalid" in sailbot. Let me know if that's unclear.

Pseudocode:


state = sailbot.getCurrentState()  
# This currently just gets current state.
# Now can get state, checks if goal latlon is invalid. If valid, then just use that. If not valid, then check if was previously invalid. If so, use that previous invalid latlon. If not previously invalid, then move global waypoint until valid and store this.

Can be used in local path visualizer as well to make both consistent.

tylerlum (commented):
Great changes, but have a new idea of how to consolidate this.

tylerlum (approved):
Beautiful code. Makes me want to cry :')

tylerlum (commented):
One other note:

I run `roslaunch local_pathfinding all.launch initial_speedup:=100 random_seed:=1234` then 

`rostopic pub /new_boats sailbot_msg/AISShip "{ID: 0, lat: 48.298, lon: -125.321, headingDegrees: 0.0, speedKmph: 10.0}"`

Then I see this

image


Essentially, the state has been updated so that the global waypoint is in a valid position, but the path still goes to the old end point. However, the local path has not changed (neither condition for createLocalPath() has been met). Therefore, it shows this until it reaches a waypoint. I think this behavior is fine for now, since it will fix itself on the next run.

patrick-5546:
A corner case that I can think of is that after the global waypoint is moved, another boat moves faster than expected and is now predicted to collide with sailbot at the global waypoint. Would we have to deal with this case?

tylerlum:
image

Found another bug. If sailbot is quite close to final waypoint, if a boat goes on top of it, the global waypoint moves 5 km back, so sailbot starts going backwards. Might make more sense to move the global waypoint AWAY from Sailbot

tylerlum:
Actually, probably better to move the global waypoint NOT on the line between sailbot and global waypoint, but ON the line between global waypoint and the global waypoint after this one. That ensures we always move forward.

tylerlum:
I tested your corner case, it seems to work. It seems like it moves back a second time. What I did was keep rostopic pub -ing more boats onto each global waypoint. Is that what you mean?

python/Sailbot.py
```diff
@@ -121,8 +126,28 @@ def getCurrentState(self):
...
             self.globalPathIndex = len(self.globalPath) - 1
 
-        return BoatState(self.globalPath[self.globalPathIndex], self.position, self.measuredWindDirectionDegrees,
-                         self.measuredWindSpeedKmph, self.AISData, self.headingDegrees, self.speedKmph)
+        state = BoatState(self.globalPath[self.globalPathIndex], self.position, self.measuredWindDirectionDegrees,
+                          self.measuredWindSpeedKmph, self.AISData, self.headingDegrees, self.speedKmph)
+
+        global goalWasInvalid
+        global movedGlobalWaypoint
+        if goalWasInvalid or not his.checkGoalValidity(state):
```
tylerlum, line 134: if not goalwasinvalid and not checkgoalvalidity:
   goalwasinvalid = true

if goalwasinvalid
   state.globalwaypoint = movedglobalwaypoint

tylerlum (approved):

PR #192, opened 2021-03-22
Add real ships from exactAIS

haakov:
Gathers ships from the 0.4x0.4 degree latlon box surrounding the sailbot every two minutes. This code works, but is in need of some polishing. Maybe the simulated ships and real ships should be in separate files/nodes?

Run with:
`roslaunch local_pathfinding all.launch use_real_ships:=True AIS_token:=<INSERT TOKEN HERE>`


haakov:
Here is a screenshot: (3 ships)
DeepinScreenshot_matplotlib_20210418235148

The same situation, shown in MarineTraffic:
Screenshot_2021-04-18 MarineTraffic Global Ship Tracking Intelligence AIS Marine Traffic

python/MOCK_AIS.py
```diff
@@ -69,13 +71,26 @@ def make_ros_message(self):
             self.id,
             self.lat,
             self.lon,
-            self.headingDegrees,
+            (90.0 -self.headingDegrees) % 360,
```
tylerlum, line 74: Yeah I see that conversion lots of places in our code. May look into making that change you talked about :)

python/MOCK_AIS.py
```diff
@@ -146,6 +167,47 @@ def gps_callback(self, msg):
     def speedup_callback(self, msg):
         self.speedup = msg.data
 
+    def get_real_ships(self):
+        lat = self.sailbot_lat
+        lon = self.sailbot_lon
+        space = "%20"
+        pos_list = ""
+        pos_list += str(lon-0.2) + space + str(lat-0.2) + space # bot left
```
tylerlum, line 175: This is a great start, but may be best to use a fixed distance like 30km x 30km square, then convert to lat/lon using our utilities functions

python/MOCK_AIS.py
```diff
@@ -146,6 +167,47 @@ def gps_callback(self, msg):
...
+        pos_list += str(lon-0.2) + space + str(lat+0.2) + space # top left
+        pos_list += str(lon+0.2) + space + str(lat+0.2) + space # top right
+        pos_list += str(lon+0.2) + space + str(lat-0.2) + space # bot right
+        pos_list += str(lon-0.2) + space + str(lat-0.2)         # bot left
+
+
+        url = 'https://services.exactearth.com/gws/wms?service=WFS&version=1.1.0&authKey=' + self.token
```
tylerlum, line 182: This looks great! Once I get the API key, I will test it out.

python/MOCK_AIS.py
```diff
@@ -155,7 +217,17 @@ def speedup_callback(self, msg):
     r = rospy.Rate(1.0 / ais_env.publishPeriodSeconds)  # hz
 
     while not rospy.is_shutdown():
+        #rospy.loginfo(ais_env.real_ships)
+        if ais_env.use_real_ships:
```
tylerlum, line 221: Agreed it may be best to have a different file for the real boats to keep it clean. Can add an `if` in the launch based on the arguments to use either fake or real

tylerlum (commented):
Great start! Will test it out soon and add more feedback. Probably good idea to put into differnet .py file and use `if` or `unless` in launch files: http://wiki.ros.org/roslaunch/XML#if_and_unless_attributes

tylerlum:
Just tried it out and it works great! Once you finish the clean up, msg me on slack and I can give final review.

tylerlum:
Fixed merge conflicts

tylerlum:
Closing in favor of #225

PR #195, opened 2021-04-08
Wandb log

tylerlum:
The purpose of this PR is to improve the logging of our current system. Previously, we had ROS nodes that stored data and created matplotlib plots when the run ended. However, this was fragile. Sometimes the plots didn't come out right, and we didn't have a nice way to zoom into the data.

Now, I have introduced wandb integration, which allows for interactive plots and easy customization of the visualization. This can significantly improve our ability to reproduce issues and improve our system.

https://wandb.ai/site
Log in with UBC Sailbot gmail (DM in Slack for the credentials).

Will store the data in interactive plots like this.
image

### DIFF
--- a/launch/all.launch
+++ b/launch/all.launch
@@ -41,8 +41,5 @@
 
     <node pkg="local_pathfinding" type="main_loop.py" name="main_loop" output="$(arg main_loop_output)"/>
     <node pkg="local_pathfinding" type="local_path_visualizer.py" name="local_path_visualizer" />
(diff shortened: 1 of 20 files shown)


tylerlum:
List out other things to log:
* Cost function weights
* Obstacle projection distance and extend time
* Sailbot max speed, etc.

IDEA: Put all constants into a config.py file that all other code can import. Then all constants will be in one place to put into wandb.

tylerlum:
Need to log in with wandb login info on Slack. General code still works without logging in, just will not log data to wandb

PR #196, opened 2021-04-10
Change from desiredHeading to heading_degrees

tylerlum:
Rostopic name change requested from Brielle.

Changed from "desiredHeading" to "heading_degrees". This refers to the heading of the boat that the local pathfinding wants the boat to be at.


briellelaw (approved):
lgtm. thank you!

PR #203, opened 2021-05-29
Update ros interface node

### LINKED ISSUE #199: Update and test ROS interface node
1. the sensor.msg type has been updated, so we should update our code to reflect this
2. We should setup a mock sensor output to actually pass sensor values through this interface and make sure the data flow is working

patrick-5546:
Resolves #199 

Also updates some setup files

### DIFF
--- a/python/ros_interface.py
+++ b/python/ros_interface.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 import rospy
-from sailbot_msg.msg import Sensors, windSensor, GPS  # does the sailbot_msg name need to be updated?
+from sailbot_msg.msg import Sensors, windSensor, GPS
+from utilities import bearingToHeadingDegrees
 
 # Constants
 CHECK_PERIOD_SECONDS = 0.1  # How often fields are updated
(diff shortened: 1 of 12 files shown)


python/ros_interface.py
```diff
@@ -1,17 +1,19 @@
 import rospy
-from sailbot_msg.msg import Sensors, windSensor, GPS
+from sailbot_msg.msg import Sensors, windSensor, GPS  # does the sailbot_msg name need to be updated?
```
tylerlum, line 2: Good question. Unfortunately, I think that it needs to be `sailbot_msg`.

tylerlum: Defined in `package.xml`, can't use `-`

tylerlum (changes requested):
Looks good, next step to add MOCK

tylerlum:
image

image

Now that the interface node is written, we need to test it. We can do this by sending in inputs and checking if the outputs match what we desire.

To run the ros interface node, we can:

`roscore` in one terminal to setup ROS
`rosrun local_pathfinding ros_interface.py` to run the interface node in another terminal

`rostopic pub /sensors <TAB>` to publish to the `sensors` topic, and see if the interface does the right thing

`rostopic echo /GPS` to see what the ros_interface node is publishing and see if it matches.

The better long term solution is to write a MOCK_sensors.py that does this process for us:

`rosrun local_pathfinding MOCK_sensors.py` (make sure to `chmod +x MOCK_sensors.py` first)

patrick-5546:
- [ ] GPS sensors - use can, ais, or both? which takes priority?
- [ ] Coordinates - always DDMM.MMMMMMM? Fix logic if not

tylerlum:
```
#!/usr/bin/env python
import rospy
import json
import time
from utilities import MAUI_LATLON

from std_msgs.msg import Float64
import sailbot_msg.msg as msg
from geopy.distance import distance

# Constants for this class
PUBLISH_PERIOD_SECONDS = 10.0  # Can keep high to simulate real global pathfinding
NEW_GLOBAL_PATH_PERIOD_SECONDS = 3600.0 * 24
AVG_WAYPOINT_DISTANCE_KM = 30  # TODO: Set this to match global pathfinding

# Global variables for tracking boat position
desiredHeading = None


def desiredHeadingCallback(data):
    global desiredHeading
    desiredHeading = data.heading_degrees

def MOCK_sensors():
    global desiredHeading

    rospy.init_node('MOCK_sensors', anonymous=True)
    pub = rospy.Publisher("sensors", msg.Sensors, queue_size=4)
    rospy.Subscrbier('heading_degrees', msg.heading, desiredHeadingCallback)
    r = rospy.Rate(float(1) / PUBLISH_PERIOD_SECONDS)  # Hz

    # Publish periodically
        sensors = msg.Sensors()
        sensors.wind = utils.globaltomeasured(speed, dir)
        pub.publish(sensors)
        r.sleep()
        lat+=1*speed*cos(desiredhheading)
        lon+=1 *speed*cos(desiredhheading)


if __name__ == '__main__':
    MOCK_sensors()
```

Rough draft of MOCK_sensors.pu

tylerlum (approved):

PR #207, opened 2021-06-20
Prevent Large Turns to New Path

### LINKED ISSUE #206: Don't change path if turn too much
Pseudocode for new change
```
                # LATLONS
                localPath.getNextWaypoint()
                _localPath.getNextWaypoint()
                state.position

                # XY
                latlonttoxy(above, state.positoin)
                angle(position, new)
                angle(position, old)
                abs(angle Careful of signs)

                if newPathCost < currentPathCost and newPathReachesGoal and abs(angle(localPath.getnextawypaotihn, state.GPS)  - _lcoalPath.firstHeading) < 10:
 ```

### PR DESCRIPTION
**Pull request by patrick-5546**
Resolves #206 

- Added `SMALL_TURN_DEGREES` constant
- Added `angleDegreesBetweenLatlons` function in utilities.py
- Added `smallTurnToNewPath` check when determining whether to update to new local path

### DIFF
--- a/python/main_loop.py
+++ b/python/main_loop.py
@@ -193,13 +193,12 @@ if __name__ == '__main__':
                 currentPathCost = localPath.getCost()
                 newPathCost = _localPath.getCost()
                 newPathReachesGoal = _localPath.reachesGoalLatlon(state.globalWaypoint)
-                smallTurnToNewPath = utils.angleDegreesBetweenLatlons(localPath.getNextWaypoint(),
-                                                                      _localPath.getNextWaypoint(), state.position) \
-                    < SMALL_TURN_DEGREES
+                absAngleToNewPath = utils.absAngleDegreesBetweenLatlons(localPath.getNextWaypoint(),
+                                                                        _localPath.getNextWaypoint(), state.position)
                 rospy.loginfo("currentPathCost = {}. newPathCost = {}. newPathReachesGoal = {}. "
-                              "smallTurnToNewPath = {}.".format(currentPathCost, newPathCost, newPathReachesGoal,
-                                                                smallTurnToNewPath))
-                if newPathCost < currentPathCost and newPathReachesGoal and smallTurnToNewPath:
+                              "absAngleToNewPath = {}.".format(currentPathCost, newPathCost, newPathReachesGoal,
+                                                               absAngleToNewPath))
+                if newPathCost < currentPathCost and newPathReachesGoal and absAngleToNewPath < SMALL_TURN_DEGREES:
                     rospy.loginfo("Updating to new local path")
                     localPath = _localPath
                 else:
--- a/python/main_loop.py
+++ b/python/main_loop.py
@@ -12,6 +12,7 @@ import matplotlib.pyplot as plt
 
 # Constants
 MAIN_LOOP_PERIOD_SECONDS = 0.5
+SMALL_TURN_DEGREES = 10
 
 # Global variable to receive path update request messages
 localPathUpdateRequested = False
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: python/main_loop.py [hunk 1]
```diff
@@ -188,13 +189,17 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
...
 
-                # Update local path if new one is better than old AND it reaches the goal
+                # Update local path if new one is better than old AND it reaches the goal AND the turn to it is small
                 currentPathCost = localPath.getCost()
                 newPathCost = _localPath.getCost()
                 newPathReachesGoal = _localPath.reachesGoalLatlon(state.globalWaypoint)
-                rospy.loginfo("currentPathCost = {}. newPathCost = {}. newPathReachesGoal = {}."
-                              .format(currentPathCost, newPathCost, newPathReachesGoal))
-                if newPathCost < currentPathCost and newPathReachesGoal:
+                smallTurnToNewPath = utils.angleDegreesBetweenLatlons(localPath.getNextWaypoint(),
```
tylerlum, line 196: Solid implementation here

python/main_loop.py
```diff
@@ -188,13 +189,17 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
...
+                                                                      _localPath.getNextWaypoint(), state.position) \
+                    < SMALL_TURN_DEGREES
+                rospy.loginfo("currentPathCost = {}. newPathCost = {}. newPathReachesGoal = {}. "
```
tylerlum, line 199: Consider logging the angle difference here for debugging (not just 1 or 0 bool)

python/utilities.py
```diff
@@ -97,6 +97,22 @@ def takeScreenshot(return_path_to_file=False):
         return fullImagePath
 
 
+def angleDegreesBetweenLatlons(latlon1, latlon2, referenceLatlon):
+    '''Calulates the angle between the XY coordinates of two latlons, given referenceLatlon located at (0,0)
```
tylerlum, line 101: calculates

python/utilities.py
```diff
@@ -97,6 +97,22 @@ def takeScreenshot(return_path_to_file=False):
...
+
+    Args:
```
tylerlum, line 103: Great docstring

python/utilities.py
```diff
@@ -97,6 +97,22 @@ def takeScreenshot(return_path_to_file=False):
...
+        latlon1 (sailbot_msg.msg._latlon.latlon): The first latlon to calculate the degree of
+        latlon2 (sailbot_msg.msg._latlon.latlon): The second latlon to calculate the degree of
+        referenceLatlon (sailbot_msg.msg._latlon.latlon): The latlon that will be located at (0, 0)
+
+    Returns:
+        float representing the absolute angle between the two latlons in degrees
+    '''
+    x1, y1 = latlonToXY(latlon1, referenceLatlon)
+    x2, y2 = latlonToXY(latlon2, referenceLatlon)
+    return abs(math.tan(y2 / x2) - math.tan(y1 / x1)) * 180 / math.pi
```
tylerlum, line 113: I think that this has a couple of bugs:
1. tan(theta) = y/x (I think you need atan)
2. It is best to use atan2 instead of atan. atan can't tell the difference between 135 degrees (-1, 1) and -45 degrees (1, -1), atan2 can
3. Careful about wrap around. Minimal example below where raw angle diff is large but they are actually very close.
image

tylerlum: Actually just realized we have a function for this. Look at `python/planner_helpers.py`

```
def abs_angle_dist_radians(angle1_radians, angle2_radians):
```

tylerlum: Test it out on some simple examples to double check if works!

tylerlum (changes requested):

python/utilities.py
tylerlum: the degrees version is now available

patrick-5546: Thanks for the help, I don't really know what I was thinking. Luckily there was an existing function to use!

patrick-5546: Just noticed the degrees version's output is still in radians. Is this the intended behaviour?

tylerlum: No that's definitely a mistake by me haha. Could you fix that? :D

tylerlum (approved):
Looks great! Could you test it on a few runs and vary the wind? Just validate that it's working

patrick-5546:
#### Testing Commands
`roslaunch local_pathfinding all.launch initial_speedup:=50`
`rostopic pub /changeGlobalWind sailbot_msg/globalWind "directionDegrees: 45.0 speedKmph: 10.0`

#### Qualitative Observations
- Behaviour is much less erratic
- However, path changes occur very rarely, even at local waypoints, which is good if the initial path is good
- But what if sailbot needs to avoid a boat path that is directly in front of its path, requiring a dramatic turn to avoid?
- Ideas that come to mind
    - Increase `SMALL_TURN_DEGREES`
    - Allow large turn path if its cost is low enough
    - Allow large turn paths at waypoints
    - Avoid paths that require a turn through upwind
    - The behaviour is good enough as it is

tylerlum:
Really great points Patrick. This looks like a big improvement. I think we should merge this first and then experiment with your other ideas in another PR :)
Couple things:
1. Your change only affects the `generateNewLocalPathToCompare` case, which is optional. It does not affect the `mustUpdateLocalPath` case, which involves upwind/downwind sailing, obstacles on path, etc.
2. I really like your ideas about the change if big cost improvement or the ratio. Should definitely try that out in next PR. May be hard to tune, but could work very well!

PR #211, opened 2021-07-04
Clean update interface

tylerlum:
The purpose of this PR is to clean up a few miscellaneous things:

* Change `speedup` and `globalWind` to a rosparam, so less messy code/global variables are needed to update and use it (speedup had some messy code to handle some things like publishing at init)
* Replace `MOCK_wind_sensor` and `MOCK_controller_and_boat` with `MOCK_sensors`
* Clean launch files so all.launch uses launch files closer to what we will use on the big day

Now:
`rosparam set /speedup 100`
`rosparam set /global_wind_direction_degrees 70`


patrick-5546 (commented):
Looked over and tested out the new `speedup` and `globalWind` implementations, looks very clean and works great :+1: . Quantitatively observed that the visualizer shuts down much faster (no waiting for sigterm) on CTRL+C.

tylerlum:
Awesome! I'll merge it in now :)

PR #216, opened 2021-07-10
Fix collision logging

JoshuaJSam:
Fixed the collision logging in collisionchecker.py. The previous code only updated the times_collided variable locally inside the method which is why the number of collisions was not logged correctly.


tylerlum:
Rejecting this, even though it's awesome

PR #217, opened 2021-07-10
Fixed the collision logging

JoshuaJSam:
Fixed the collision logging in collisionchecker.py. The previous code only updated the times_collided variable locally inside the method which is why the number of collisions was not logged correctly.


tylerlum (approved):
Looks great and works as expected. Merging now.

PR #224, opened 2021-07-17
Refactor MOCK AIS

### LINKED ISSUE #223: Clean MOCK_AIS
Some things to be cleaned:
* class RandomShip and Ship and RealShip should be put together into one Ship class that simply stores data, since they should store the same data. 
* Can make subclass moving ship that adds the move() function. Ship should not need both publish period seconds and speedup, but instead should have move(time_since_last_update), since we only care about publishPeriod * speedup = time_since_last_update
* User of the class can pass in the random position coordinates to get the same behavior as RandomShip

tylerlum:
Resolves #223


tylerlum:
Tested this pretty thoroughly and things are working as expected. Going to merge this in quickly so that other PRs can build on top of it.

PR #225, opened 2021-07-17
Exact ais updated

tylerlum:
@haakov's PR #192 , but rebased to new MOCK_AIS, which has been refactored.


haakov:
@tylerlum ready for review. I've removed the `use_real_ships` argument, exactAIS will now be used if an AIS token is given, random simulated ships are used otherwise.

Run:
`roslaunch local_pathfinding all.launch AIS_token:=<INSERT TOKEN HERE>`

tylerlum (commented):
image

lol that I can't approve.

Really nice clean code. I like the separation between the MOCK_AIS and MOCK_exactAIS and the use the `if` in roslaunch. Tested out with and without exactAIS and works perfectly. Going to merge in right away so we can start using it more

PR #229, opened 2021-07-30
Visualizer starts right away, but without ships or global waypoint

### LINKED ISSUE #218: Start local path visualizer right away
Right now, local path visualizer waits until the first run of main loop to show anything. This can take some time and won't show us what's going on in the first run. Would be best to allow it to show whatever it has and update over time.

patrick-5546:
Fixes #218

### DIFF
--- a/python/local_path_visualizer.py
+++ b/python/local_path_visualizer.py
@@ -195,7 +195,7 @@ if __name__ == '__main__':
     # Making the tail half of the visualizer
     positionPlotTail, = axes.plot(positionXY[0] - 1 * math.cos(math.radians(state.headingDegrees)),
                                   positionXY[1] - 1 * math.sin(math.radians(state.headingDegrees)),
-                                  marker=(5, 0, state.headingDegrees - 90), color='r', markersize=.7 * markersize)
+                                  marker=(5, 0, state.headingDegrees - 90), color='r', markersize=0.7 * markersize)
 
     # Setup plot xy limits and labels
     axes.set_xlim(xNLim, xPLim)
@@ -272,6 +272,7 @@ if __name__ == '__main__':
             if 'nextGlobalWaypointPlot' in globals():
                 nextGlobalWaypointPlot.set_xdata(nextGlobalWaypointXY[0])
                 nextGlobalWaypointPlot.set_ydata(nextGlobalWaypointXY[1])
+                nextGlobalWaypointPlot.set_markersize(markersize)
                 globalWaypointReachedPlot.set_xdata(glob_X)
                 globalWaypointReachedPlot.set_ydata(glob_Y)
             else:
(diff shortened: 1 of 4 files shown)


python/local_path_visualizer.py
```diff
@@ -221,63 +220,104 @@ def y(x):
...
-        globalWaypointReachedPlot.set_xdata(glob_X)
-        globalWaypointReachedPlot.set_ydata(glob_Y)
-
-        # Resize axes if needed
-        if needAxesResized(positionXY, nextGlobalWaypointXY, xPLim, xNLim, yPLim, yNLim):
-            xPLim, xNLim, yPLim, yNLim = getXYLimits(positionXY, nextGlobalWaypointXY)
-            axes.set_xlim(xNLim, xPLim)
-            axes.set_ylim(yNLim, yPLim)
+        if nextGlobalWaypoint:
+            referenceLatlon = nextGlobalWaypoint  # Matches createLocalPathSS referenceLatlon for best results
```
patrick-5546, line 244: Currently, the plot the appears right away looks like this
image
Since `shipsXY` uses `referenceLatlon`, which is equal  to `nextGlobalWaypoint`, the ships do not appear in the initial plot. Is this the desired behavior?

tylerlum:
Great stuff! You made an important observation. Our refernence latlon (where xy=(0,0)) is at the global waypoint latlon. With this in mind, it might make more sense to make the reference latlon at the sailbot location. Could you test that out?

tylerlum:
```
➜  local-pathfinding git:(SetStartGoalLatlon) ✗ git grep "referenceLatlon = "
```

image

Try changing all places with global waypoint or next goal latlon etc. to the sailbot position latlon. Let me know if that makes sense or not.

patrick-5546:
The boats are visualized now even when there are no solutions.
image

patrick-5546:
However, since sailbot is plotted after the local path, the local path is on top of sailbot
image

tylerlum:
Thanks for the update. That's annoying. Could you try this? https://matplotlib.org/3.1.1/gallery/misc/zorder_demo.html if too hard, this is fine though!

patrick-5546:
One side effect of using state.position as referenceLatlon is that the previous view button will not work as intended.
Peek 2021-08-07 10-41

tylerlum:
Thanks for all the hard work. Didn't think about that being a problem :( let's think about it some more to see if this is really worth doing, since the continuously moving reference is not very nice

tylerlum:
The only thing missing for visualizer is the global waypoint

Can add this here in main_loop.py
```
    # Create first path and track time of updates
    state = sailbot.getCurrentState()
    nextGlobalWaypointPublisher.publish(state.globalWaypoint)  # ADD THIS!

    # No prev path to fall back on, so give lots of time
    localPath, lastTimePathCreated = createNewLocalPath(sailbot, float('inf'), desiredHeadingPublisher, None)
    desiredHeadingDegrees = utils.getDesiredHeadingDegrees(state.position, localPath.getNextWaypoint())
    sailbot.newGlobalPathReceived = False
```

patrick-5546:
Case with too many boats works as expected
image

tylerlum:
This picture makes me want to cry. Can't decide if in joy or sadness.

tylerlum:
This picture makes me want to cry. Can't decide if in joy or sadness.

tylerlum (approved):
Tried it and it works well!

PR #231, opened 2021-08-11
Remove global wind node

tylerlum:
Run the usual: `roslaunch local_pathfinding all.launch` check everything still working. Also updated README img.


JoshuaJSam:
Looks good, and is working as expected

PR #234, opened 2021-08-26
Log AIS Ships

### LINKED ISSUE #233: Continuously log and store AIS data, then perform exploratory data analysis on it
* Figure out if feasible (how much storage this would take)
* Look at data (how much boats disappear, change speed, change heading, match where we expected them to be, boat sizes, etc.)

patrick-5546:
Resolves #233 

1. Creates a log file for AIS ships every time `get_real_ships()` is called in the directory `ais_ships_log/`
    - Turn on logging by adding `log_exactais:=true` to the roslaunch command
2. Simple analysis of exactAIS data
    - Note: the headings saved from exactAIS data are misleadingly named: they are NOT the same units as headingDegrees in the pathfinding code. To convert to heading degrees, use the `bearingToHeadingDegrees` function in utilities.py


tylerlum (changes requested):
Looks good but a few suggestions:

exactais_scraper.py and exactais_ships_analysis.py look very independent of all other parts of this code. Could you put these in their own folder, probably not inside python directory but outside. Something like `exactais_analysis`. Since this is not pathfinding code, we can keep it away. Same with heatmap.txt put in there too

Also, add a rosparam for the logging. May take up a lot of space to store at all times.

patrick-5546:
Example of bounding boxes in the scraped exactAIS data for `dimension=5`
image

patrick-5546:
Quick observations from tracking boats in bounding boxes *(n,n)* for *n* in [0,4]:
- Boat behavior does not seem to be dependent on where it is located in the ocean (aka which bounding box it is in)
- There are different types of boats that have different behaviors
    - Some boats maintain a constant speed (±1 knot) and heading (±5 degrees): bulk carrier
        - Some reported the exact same speed and heading for 8 hours
    - Some boats vary greatly in speed and heading, but are very slow in general, like a research vessel

tylerlum:
Awesome, can you put some of the speed and heading plots here? Would be great to have it recorded and also potentially model the research vessels :)

python/MOCK_exactAIS.py
```diff
@@ -80,16 +81,40 @@ def get_real_ships(self):
...
+                    'speedKmph': float(real_ship[u'sog']) * 0.54
+                },
+                'length': real_ship[u'length'],
+                'width': real_ship[u'width']
+            }
+            self.ais_ships_log.append(log_ship)
+        self.log_ais_ships()
+
+    def log_ais_ships(self):
+        now = datetime.datetime.now()
```
tylerlum, line 111: Add a check like rosparam get here and in launch file, then good to merge!

tylerlum (approved):

patrick-5546:
> Awesome, can you put some of the speed and heading plots here? Would be great to have it recorded and also potentially model the research vessels :)

https://github.com/UBCSailbot/local-pathfinding/blob/e20a63ef13daa1c50326c9953e96ef8c2a1cefa2/exactais_analysis/boat_tracker_data/bb_33_mmsi_538005923.png

PR #235, opened 2021-09-11
Add Noise to GPS and Wind sensors

patrick-5546:
Draw a random sample from a Gaussian distribution using [numpy.random.normal](https://numpy.org/doc/stable/reference/random/generated/numpy.random.normal.html). GPS sensors have a much lower standard deviation than wind sensors (and wind speed itself varies quite a bit). The aim was for GPS sensors to vary ~10-20m, but further refinement of the standard deviation constants may be necessary.

Add noise to sensors with `sensor_noise:=true` in the roslaunch command.

### DIFF
--- a/python/MOCK_sensors.py
+++ b/python/MOCK_sensors.py
@@ -37,16 +37,9 @@ def getStartLatLon(defaultLat, defaultLon):
     return (startLat, startLon)
 
 
-def getSensorNoise():
-    # Read in sensor noise
-    sensorNoise = rospy.get_param('sensor_noise', default=False)
-    sensorNoise = bool(sensorNoise)
-    return sensorNoise
-
-
 class MOCK_SensorManager:
     def __init__(self, startLat, startLon, startHeadingDegrees, startSpeedKmph, startGlobalWindSpeedKmph,
-                 startGlobalWindDirectionDegrees, sensorNoise):
+                 startGlobalWindDirectionDegrees):
         # Initialize starting values
         self.lat = startLat
         self.lon = startLon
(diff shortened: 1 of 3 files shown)


python/MOCK_sensors.py
```diff
@@ -88,33 +91,41 @@ def update(self):
...
         # data.gps_can_magnetic_variation_degrees
         # data.gps_can_state
 
         # data.gps_ais_timestamp_utc
-        data.gps_ais_latitude_degrees = self.lat
-        data.gps_ais_longitude_degrees = self.lon
-        data.gps_ais_groundspeed_knots = self.speedKmph / KNOTS_TO_KMPH
-        data.gps_ais_track_made_good_degrees = headingToBearingDegrees(self.headingDegrees)
-        data.gps_ais_true_heading_degrees = headingToBearingDegrees(self.headingDegrees)
+        data.gps_ais_latitude_degrees = np.random.normal(self.lat, STDEV_GPS, 1)[0]
```
tylerlum, line 122: I believe you can just do `np.random.normal(self.lat, STDEV_GPS)`, no need for the  1 and [0].

tylerlum (changes requested):
I like it so far but 2 suggestions:

1. I think it's best to actually keep the publish() function almost the same as before (exact values). Then before the self.publisher.publish(data) line, you can run something like data = self.add_noise(data). This decouples the usual stuff from the noise stuff
2. Could you add a rosparam and argument so we can turn it on and off. Eg. `sensor_noise:=true`. The above change will make this easier.

python/MOCK_sensors.py
```diff
@@ -37,9 +37,16 @@ def getStartLatLon(defaultLat, defaultLon):
     return (startLat, startLon)
 
 
+def getSensorNoise():
+    # Read in sensor noise
+    sensorNoise = rospy.get_param('sensor_noise', default=False)
+    sensorNoise = bool(sensorNoise)
```
tylerlum, line 43: Double check logic about string to bool. Right now bool("false") = True lol

python/MOCK_sensors.py
```diff
@@ -142,8 +142,56 @@ def publish(self):
         # data.gyroscope_y_velocity_millidegreesps
         # data.gyroscope_z_velocity_millidegreesps
 
+        if self.sensorNoise:
```
tylerlum, line 145: Right now, just looks at sensor_noise param from beginning. It would be nice if mid sim, we could run `rosparam set sensor_noise false` and it turn off.

`if getSensorNoise():`

tylerlum (changes requested):

tylerlum (approved):

Issue #213, opened 2021-07-07
Dynamically update number of ais ships


tylerlum:
Would be great to run `rosparam set /num_ais_ships 30` to add a bunch of boats mid sim.
Also `rosparam set /num_ais_ships 2` to get rid of most boats.

tylerlum:
A good example of rosparam usage is in https://github.com/UBCSailbot/local-pathfinding/blob/master/python/MOCK_sensors.py

tylerlum:
Right now the pseudocode looks like:

```
num_boats = rospy.get_param('num_ais_ships', default=5)
boats = [create_boat() for i in range(num_boats)]
every 1s:
    for boat in boats:
        boat.move_forward()
```

tylerlum:
But it would be best to look like:
```
num_boats = rospy.get_param('num_ais_ships', default=5)
boats = [create_boat() for i in range(num_boats)]
every 1s:
    ## NEW PART BEGIN
    num_boats = rospy.get_param('num_ais_ships', default=5)
    # Create new boats
    if num_boats > len(boats):
        num_new_boats = num_boats - len(boats)
        for i in range(num_new_boats):
            boats.append(create_boat())
    # Remove some boats
    elif num_boats < len(boats):
        num_removed_boats = len(boats) - num_boats
        for i in range(num_removed_boats):
            boats.pop()
    else:
        # Same number of boats, so no need to do anything different
    ## NEW PART END
    for boat in boats:
        boat.move_forward()
```

tylerlum:
Current code looks like:
```
if __name__ == '__main__':
    # Get ais_file parameter
    ais_file = rospy.get_param('ais_file', default=None)
    ais_env = MOCK_AISEnvironment(PORT_RENFREW_LATLON.lat, PORT_RENFREW_LATLON.lon, ais_file)
    r = rospy.Rate(1.0 / ais_env.publishPeriodSeconds)  # hz
    while not rospy.is_shutdown():
        data = ais_env.make_ros_message()
        ais_env.move_ships()
        ais_env.publisher.publish(data)
        r.sleep()
```

tylerlum:
I think we could add something like `ais_env.check_new_num_ship()` before `ais_env.move_ships()`

`class MOCK_AISEnvironment` has `self.ships = []` that matches the boat list above

PR #236, opened 2021-09-18
Dynamically update number of ais ships

JoshuaJSam:
This PR adds the ability to dynamically update number of AIS ships

#213

### DIFF
--- a/python/MOCK_AIS.py
+++ b/python/MOCK_AIS.py
@@ -140,6 +140,18 @@ class MOCK_AISEnvironment:
     def gps_callback(self, msg):
         self.sailbot_lat = msg.lat
         self.sailbot_lon = msg.lon
+    
+    def check_new_num_ship(self, previous_num_ships):
+        updated_num_ships = rospy.get_param('num_ais_ships', default=5)
+        if updated_num_ships is not previous_num_ships:
+
+            self.ships = []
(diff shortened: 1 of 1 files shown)


python/MOCK_AIS.py
```diff
@@ -140,6 +140,18 @@ def remove_boat_callback(self, msg):
     def gps_callback(self, msg):
         self.sailbot_lat = msg.lat
         self.sailbot_lon = msg.lon
+    
+    def check_new_num_ship(self, previous_num_ships):
+        updated_num_ships = rospy.get_param('num_ais_ships', default=5)
```
tylerlum, line 145: len(self.ships) is number of ships currently
updated_num_ships = rospy.get_param('num_ais_ships', default=5) updated number

if len(self.ships) == updated_num_ships: do nothing
elif len(self.ships) > updated_num_ships: num_remove = len(self.ships) - updated_num_ships. for loop remove ships from self.ships (eg. if had 10 ships but want 2 now, then remove 8 ships from list)
elif len(self.ships) < updated_num_ships: num_add = updated_num_ships - len(self.ships). for loop add ships to self.ships (eg. if had 3 ships but want 7 now, then add 4 random ships to list)

JoshuaJSam:
See #239

PR #239, opened 2021-09-25
Added feature to change number of boats during runtime

JoshuaJSam:
This PR adds the ability to dynamically update number of AIS ships

#213


JoshuaJSam:
To test:

1) Get things setup
```
source ~/catkin_ws/devel/setup.bash    
roslaunch local_pathfinding all.launch

```

2) Add boats
`rosparam set /num_ais_ships 50`

3) Remove boats (From 50 to 10)
`rosparam set /num_ais_ships 10`

tylerlum (changes requested):
PR overall logic looks great!

I ran your test and it seems to work exactly as desired :)

I also tried running a different test:

```
➜  local-pathfinding git:(dynamically_add_ships) ✗ roslaunch local_pathfinding all.launch num_ais_ships:=10
➜  local-pathfinding git:(dynamically_add_ships) ✗ rosparam set /num_ais_ships 0
```

I got this error.
image

I believe we have a `self.numShips` variable that isn't updated. Tbh, I feel like the best solution is to remove that variable entirely and just replace any use of `self.numShips` with `len(self.ships)`, since that removes a state variable and removes a potential bug source. Let me know what you think.

tylerlum:
```
        for i in range(self.numShips):
            self.ships[i].move(movement_time_seconds)
```
Should just be range(len(self.ships))

python/MOCK_AIS.py
```diff
@@ -111,46 +134,85 @@ def set_random_seed(self):
...
+                "randomSeed = {}. Not setting seed".format(randomSeed))
 
     def move_ships(self):
         speedup = rospy.get_param('speedup', default=1.0)
         movement_time_seconds = speedup * self.publishPeriodSeconds
 
-        for i in range(self.numShips):
-            self.ships[i].move(movement_time_seconds)
+        
+        for i in (i for i in range(len(self.ships)) if self.ships):
```
JoshuaJSam, line 145: Is this line too verbose? I thought it would be cleaner to use a generator expression instead of this, both are ok to me:
`
for i in range(len(self.ships))
    if self.ships
       #rest of code here


`

PR #246, opened 2021-10-11
Formats and prints path latitudes and longitudes to logs

jamenkaye:
Please let me know if my code is messy, over complicated, or too difficult to follow, or could use any other further work. The code here could be replaced with
`rospy.loginfo("path is {}".format(localPath.getLatlons()))`,
but this causes each line in the resulting log to contain the longitude of the previous point and the latitude of the current point.

### DIFF
--- a/python/main_loop.py
+++ b/python/main_loop.py
@@ -13,6 +13,7 @@ import matplotlib.pyplot as plt
 # Constants
 MAIN_LOOP_PERIOD_SECONDS = 0.5
 SMALL_TURN_DEGREES = 10
+LOGGED_LATLONS_PER_LINE = 3
 
 # Global variable to receive path update request messages
 localPathUpdateRequested = False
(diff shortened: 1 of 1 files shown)


tylerlum:
Great work Jamen and great review Patrick. Couple of other notes:

image
This shows your code builds and passes out standard tests (doesn't break things) :D. It also says there's a lint error, which is just a style issue. You can click details, login with github, then scroll down

image

This shows the issue. 

```
/catkin_ws/src/local-pathfinding/python/main_loop.py:220:9: E303 too many blank lines (2)
/catkin_ws/src/local-pathfinding/python/main_loop.py:226:21: E117 over-indented
/catkin_ws/src/local-pathfinding/python/main_loop.py:238:9: E265 block comment should start with '# '
```
A few style things to improve on lines 220, 226 and 238 of the main loop :)

tylerlum:
image

I see Patrick's point, but I like this too. More of a style/preference thing. 

The code is pretty clear and easy to understand. The only part I think you could work on is the j = 14, which isn't very clear. See if it's possible to do something like.

```
        for latlon in localPath.getLatlons():
            latLonsToLog += "Lat: {}, lon {}".format(latlon.lat, latlon.lon)
```

python/main_loop.py
```diff
@@ -216,6 +217,22 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
         # Publish local path
         localPathPublisher.publish(msg.path(localPath.getLatlons()))
 
+        # Format and log local path
+        latLonsToLog = "Path latlons are:"
+
+        for i in range(len(localPath.getLatlons())):
+            if(i % LOGGED_LATLONS_PER_LINE == 0):
```
tylerlum, line 224: Great use of all caps named constant :)

python/main_loop.py
```diff
@@ -216,6 +217,22 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
...
+                latLonsToLog += "\n"
+
+            latLon = localPath.getLatlons()[i]
+            latLonsToLog += "(lat: {}, lon: {})".format(latLon.lat, latLon.lon)
```
tylerlum, line 228: Looks very clear and easy to understand
image

tylerlum (approved):
Approved! Great work!
Feel free to click this Squash and Merge button to merge it in
image

patrick-5546 (approved):

PR #248, opened 2021-10-11
ROS Interface Logging

### LINKED ISSUE #242: NEW MEMBER TASK: Add detailed debugging logs for ros_interface.py
Related to #241. When we are testing the boat on the water, many things can go wrong. In particular, we expect issues with reading in sensor data and handling noise.

Right now, our sensor processing is handled entirely in `python/ros_interface.py`, which takes in sensor data and converts it into meaningful summaries to be used by pathfinding. This works properly with MOCK_sensors, but may not work with real sensors. 

To prepare for on water testing, we need to have a high-quality method of logging to understand what is going wrong. Essentially, we should add a ton of logging statements into `python/ros_interface.py`. Eg. `rospy.logdebug("Received wind sensor data {}".format(new_wind))`. This is like a print statement with two differences:
1. It only prints if we are in debug mode. This way, we can be very "verbose" if we want, or prints less if we only want essential info (eg. failures, errors, warnings)
2. It logs the text to a log file (typically in `~/.ros/log`) so it can be reviewed later. It has a debug tag on the text so we know which ones are working.

Add this in many places. Eg. 
```
        else:
# print all trusted values, print sum, print length, print this final value
            return float(sum(trustedVals)) / len(trustedVals)
```

Details about how different logging levels work
http://wiki.ros.org/rospy/Overview/Logging

patrick-5546:
Resolves #242.

In `~/.ros/log/latest/ros_interface-3.log`:

Example of normal operation:
```
[rosout][INFO] 2021-10-11 12:33:51,126: Getting trusted average of wind direction
[rosout][INFO] 2021-10-11 12:33:51,128: Trusted average: -261 / 3 = -87.0
```

Example of warning (warnings will also show up in stdout):
```
[rosout][INFO] 2021-10-11 12:33:51,031: Getting trusted average of wind direction
[rosout][WARNING] 2021-10-11 12:33:51,033: pastValue is None or 0, using only current sensor data
[rosout][INFO] 2021-10-11 12:33:51,034: Trusted average: -261 / 3 = -87.0
```

### DIFF
--- a/python/ros_interface.py
+++ b/python/ros_interface.py
@@ -122,7 +122,8 @@ class RosInterface:
             return vals[0]
         else:
             truAvg = float(sum(trustedVals)) / len(trustedVals)
-            rospy.loginfo('Trusted average: {} / {} = {}'.format(sum(trustedVals), len(trustedVals), truAvg))
+            rospy.loginfo('Trusted average: (sum{} = {}) / {} = {}'
+                          .format(trustedVals, sum(trustedVals), len(trustedVals), truAvg))
             return truAvg
 
     def getTrusted(self, pastValue, vals):
@@ -134,15 +135,16 @@ class RosInterface:
             rospy.logwarn('pastValue is None or 0, using only current sensor data')
             return vals
 
+        rospy.loginfo('pastValue = {}'.format(pastValue))
         trustedVals = []
         for i in range(len(vals)):
             value = vals[i]
             if value is not None and abs(float(pastValue - value) / pastValue) < MAX_ALLOWABLE_PERCENT_ERROR:
                 trustedVals.append(value)
             elif value is None:
-                rospy.logwarn('Sensor {} is None, skipping'.format(i))
+                rospy.logwarn('Sensor {} data is None, skipping'.format(i))
             else:
-                rospy.logwarn('Sensor {} is very different from pastValue, skipping'.format(i))
+                rospy.logwarn('Sensor {} data = {}, which is very different from pastValue; skipping'.format(i, value))
         return trustedVals
 
     def get_global_wind(self):
--- a/python/ros_interface.py
+++ b/python/ros_interface.py
@@ -80,27 +80,33 @@ class RosInterface:
         self.past_measured_wind_direction = self.measured_wind_direction
 
         # GPS sensors - use can, ais, or both? which takes priority? true heading / track made good usage?
+        rospy.loginfo('Getting trusted average of GPS latitude')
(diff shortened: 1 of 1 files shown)


python/ros_interface.py
```diff
@@ -80,27 +80,33 @@ def translate(self):
         self.past_measured_wind_direction = self.measured_wind_direction
 
         # GPS sensors - use can, ais, or both? which takes priority? true heading / track made good usage?
+        rospy.loginfo('Getting trusted average of GPS latitude')
         self.gps_lat_decimalDegrees = self.getTrustedAvg(
                 pastValue=self.past_gps_lat_decimalDegrees,
                 vals=[data.gps_can_latitude_degrees, data.gps_ais_latitude_degrees])
+        rospy.loginfo('Getting trusted average of GPS longitude')
```
tylerlum, line 87: Great work with this, these make it very clear

python/ros_interface.py
```diff
@@ -111,23 +117,32 @@ def getTrustedAvg(self, pastValue, vals):
...
         trustedVals = self.getTrusted(pastValue, vals)
         if len(trustedVals) == 0:
-            rospy.logwarn("Values don't match with pastValue, or if pastValue is None or 0, defaulting to first sensor")
+            rospy.logwarn('No trusted values, defaulting to sensor 0')
+            rospy.loginfo('Trusted average: {}'.format(vals[0]))
             return vals[0]
         else:
-            return float(sum(trustedVals)) / len(trustedVals)
+            truAvg = float(sum(trustedVals)) / len(trustedVals)
+            rospy.loginfo('Trusted average: {} / {} = {}'.format(sum(trustedVals), len(trustedVals), truAvg))
```
tylerlum, line 125: Could you also print the actual list of values? 
Eg. Trusted average of [1, 3, 4] = 8 / 3 = 2.66666
Would be nice to see the raw readings.

python/ros_interface.py
```diff
@@ -111,23 +117,32 @@ def getTrustedAvg(self, pastValue, vals):
...
         '''Returns vals without outlier terms like None
             - Assumes sensor readings are consistent and first sensor initial reading is accurate
             - Could modify past value to be the average of the past few outputs (store in list)
         '''
         if pastValue is None or pastValue == 0:
+            rospy.logwarn('pastValue is None or 0, using only current sensor data')
             return vals
 
         trustedVals = []
-        for value in vals:
```
tylerlum, line 128: Log what the past value is.

tylerlum (changes requested):
A few comments of extra things to log, then good to go!

PR #250, opened 2021-10-11
Improve logging, cleanup code, and fix heading coordinate conversion in handle invalid state

tylerlum:
Before had ugly unneeded warnings like:

image
image

These are when it has value "", so it uses default values. No warning is needed for these, since these are regular cases. It should show warning for bad inputs like
image (second row bad, first row good)

New code only has these yellow warnings if invalid input is passed in, not if using defaults.

image

ALSO improve logging of desired heading. Before, confusion arose between this conversion.
image

Now add clarification with verbose logging.
image

ALSO fix issue with handle invalid state. Forgot to have coordinate conversion of published angle, so this would not move us in the right direction to safety.

### DIFF
--- a/python/MOCK_AIS.py
+++ b/python/MOCK_AIS.py
@@ -4,7 +4,7 @@ import random
 import json
 
 from geopy.distance import distance
-from utilities import headingToBearingDegrees, PORT_RENFREW_LATLON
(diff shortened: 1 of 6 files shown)


patrick-5546 (approved):
Works as expected

PR #257, opened 2021-10-14
Revise Sensor Filtering Logic

### LINKED ISSUE #256: Revise Sensor Filtering Logic
### Context
- Local pathfinding uses data from 2 types of sensors: GPS and wind
    - 2 GPS sensors:  CAN and AIS
    - 3 [identical?] wind sensors
- Data from multiple sensors are filtered into one input signal to get a more consistent and accurate reading

### Current Filtering Logic
- Identical for GPS and wind
- Documented in `getTrustedAvg()` in `python/ros_interface.py`
- Issue: input signals do not always remain constant, because the boat is moving/accelerating/turning
    - Past values might not be an accurate reference for new values

#### New Implementations
1. If the current sensor data is close together, average them (or take the median); else, choose the value closest to the past value
    - Same issue as current implementation when sensor data is not close together
2. Change the reference value from `pastVal` to `m + pastVal`, where `m` is the best fit slope through the last few past values
    - Better accuracy for linear behavior
3. Combine the first 2 ideas

#### Revisions of Current Implementation
1. Make current value independent of past value
   a. Potentially only for fields that are supposed to exhibit linear behavior (i.e. GPS position when speed is nonzero)

#### General
- Stop using sensor if it is detected to be broken (same value for a long time)
- Once we have real data
    - Determine whether filtering using past value is needed for each field
    - See if one of the GPS sensors is always more consistent and accurate

patrick-5546:
Resolves #256 

Implemented Idea 1 from the linked issue

### DIFF
--- a/python/ros_interface.py
+++ b/python/ros_interface.py
@@ -2,13 +2,15 @@
 import numpy as np
 import rospy
 from sailbot_msg.msg import Sensors, windSensor, GPS, globalWind
-from utilities import bearingToHeadingDegrees, measuredWindToGlobalWind
+from utilities import bearingToHeadingDegrees, measuredWindToGlobalWind, boundTo180, absDiffBetweenAngles, averageAngles
 
 
 # Constants
 CHECK_PERIOD_SECONDS = 0.1  # How often fields are updated
 KNOTS_TO_KMPH = 1.852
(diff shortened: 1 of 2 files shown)


patrick-5546:
- [ ] Fix relative error equation

patrick-5546:
- Replaced by https://github.com/UBCSailbot/local-pathfinding/pull/283

PR #260, opened 2021-10-20
Saves target global waypoint to logs and wandb w/ cleaner print and v…

jamenkaye:
…ariable names

### DIFF
--- a/python/main_loop.py
+++ b/python/main_loop.py
@@ -239,10 +239,12 @@ if __name__ == '__main__':
         localPath.updateWindDirection(state)
         localPath.updateObstacles(state)
 
-        # Publish nextLocalWaypoint and nextGlobalWaypoint and path cost
+        # Publish nextLocalWaypoint and nextGlobalWaypoint and path cost. Log nextGlobalWaypoint
         nextLocalWaypointPublisher.publish(localPath.getNextWaypoint())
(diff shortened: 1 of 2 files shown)


python/main_loop.py
```diff
@@ -239,10 +239,12 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
         localPath.updateWindDirection(state)
         localPath.updateObstacles(state)
 
-        # Publish nextLocalWaypoint and nextGlobalWaypoint and path cost
+        # Publish nextLocalWaypoint and nextGlobalWaypoint and path cost. Log nextGlobalWaypoint
         nextLocalWaypointPublisher.publish(localPath.getNextWaypoint())
         previousLocalWaypointPublisher.publish(localPath.getPreviousWaypoint())
         nextGlobalWaypointPublisher.publish(state.globalWaypoint)
+        globalWaypointLogStr = "Next Global Waypoint is: (lat: {}, lon: {})"
```
tylerlum, line 246: The text looks great! All nice on one line. 

Can you put this into the loginfo directly? No need to keep the `globalWaypointLogStr` around

```
rospy.loginfo("Next Global Waypoint is: (lat: {}, lon: {})".format(state.globalWaypoint.lat, state.globalWaypoint.lon))
```

PR #261, opened 2021-10-21
Remove sensor logic

patrick-5546:
Remove sensor filtering logic for the first in water tests to reduce the risk of something unexpected happening.


tylerlum (approved):
Looks great. Merging now.

PR #264, opened 2021-11-20
added ability to corrupt wind sensors

jamenkaye:
- Should I write more documentation for how to use this code? Should I update the readme document?
- What is the convention for wind direction angles here? I'm assuming angles need to be on [0, 360) and would need to make a small change otherwise
- Please let me know if there are any other issues or if I should add something else!

### DIFF
--- a/python/MOCK_sensors.py
+++ b/python/MOCK_sensors.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# from re import I
 import numpy as np
 import rospy
 import random
@@ -24,12 +23,12 @@ START_GLOBAL_WIND_DIRECTION_DEGREES = 0
 START_GLOBAL_WIND_SPEED_KMPH = 10
 STDEV_GPS = 0.00001
 STDEV_WIND = 0.1
-MAX_BROKEN_WINDSPEED_KTS = 50
-MAX_ANGLE_ERROR = 60
-MAX_WINDSPEED_MULTIPLIER_ERROR = 0.5
-MAX_WINDSPEED_ADDER_ERROR = 5
-MIN_LOOPS_UNTIL_BROKEN = 30
-MAX_LOOPS_UNTIL_BROKEN = 100
+MAX_BROKEN_WINDSPEED_KNOTS = 30
+MAX_ANGLE_ERROR_DEG = 60
+MAX_WINDSPEED_MULTIPLIER_ERROR = 2  # Should be >= 1
+MAX_WINDSPEED_ADDER_ERROR_KNOTS = 5
+MIN_LOOPS_TO_BREAK = 30
+MAX_LOOPS_TO_BREAK = 100
 
 
 class MOCK_SensorManager:
@@ -45,7 +44,7 @@ class MOCK_SensorManager:
         self.measuredWindSpeedKmph, self.measuredWindDirectionDegrees = globalWindToMeasuredWind(
             self.globalWindSpeedKmph, self.globalWindDirectionDegrees, self.speedKmph, self.headingDegrees)
         self.publishPeriodSeconds = SENSORS_PUBLISH_PERIOD_SECONDS
-        self.brokenSensorData = "uninitialized"
+        self.brokenSensorData = None
 
         # Setup ROS node inputs and outputs
         self.publisher = rospy.Publisher("sensors", msg.Sensors, queue_size=4)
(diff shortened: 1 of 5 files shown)


tylerlum:
Should I write more documentation for how to use this code? Should I update the readme document?
* Good idea, could you update the README with a description of how to use it?

What is the convention for wind direction angles here? I'm assuming angles need to be on [0, 360) and would need to make a small change otherwise
* I'm not sure if angles are in [0, 360) or [-180, 180). Could you make it work for both or would that be hard?

python/MOCK_sensors.py
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
+# from re import I
```
tylerlum, line 2: small thing: remove this

tylerlum:
Clean way to throttle updates:
```
rospy.loginfo_throttle(60, "This message will print every 60 seconds")
```

python/MOCK_sensors.py
```diff
@@ -122,8 +131,151 @@ def publish(self):
 
         if rospy.get_param('sensor_noise', default=False):
             data = self.add_noise(data)
+
+        brokenWindSensorStatus = rospy.get_param('broken_wind_sensors', default="not_broken")
```
tylerlum, line 135: Would be best if you put this into another function to make the `publish` function more clear.

```
brokenWindSensorStatus = rospy.get_param('broken_wind_sensors', default="not_broken")
if brokenWindSensorStatus != "not_broken":
    # update
```

python/MOCK_sensors.py
```diff
@@ -37,6 +45,7 @@ def __init__(self, startLat, startLon, startHeadingDegrees, startSpeedKmph, star
         self.measuredWindSpeedKmph, self.measuredWindDirectionDegrees = globalWindToMeasuredWind(
             self.globalWindSpeedKmph, self.globalWindDirectionDegrees, self.speedKmph, self.headingDegrees)
         self.publishPeriodSeconds = SENSORS_PUBLISH_PERIOD_SECONDS
+        self.brokenSensorData = "uninitialized"
```
patrick-5546, line 48: More pythonic to initialize to `None`

patrick-5546: Then you can check with `if self.brokenSensorData is None`

python/MOCK_sensors.py
```diff
@@ -122,8 +131,151 @@ def publish(self):
...
+            # brokenSensorData is used to store an integer 1, 2, or 3 representing which sensor is broken
+
+            if(self.brokenSensorData == "uninitialized"):
+                # set up brokenSensorData so publish(self) remembers which sensor is stuck at zero
+                randomSeed = get_rosparam_or_default_if_invalid('random_seed', default=None, rosparam_type_cast=str)
+                random.seed(randomSeed)
+                self.brokenSensorData = random.randint(1, 3)
+
+            self.corrupt_wind_sensor(data, self.brokenSensorData, "angle", 0)
+            self.corrupt_wind_sensor(data, self.brokenSensorData, "speed", 0)
```
patrick-5546, line 150: Can combine this into one function call that does both

python/MOCK_sensors.py
```diff
@@ -122,8 +131,151 @@ def publish(self):
...
+                random.seed(randomSeed)
+
+                brokenSensorID = random.randint(1, 3)
+                theta = (2*random.random() - 1) * MAX_ANGLE_ERROR
+                windspeedMultiplier = 1 + (2*random.random() - 1) * MAX_WINDSPEED_MULTIPLIER_ERROR
+                windspeedAdder = (2*random.random() - 1) * MAX_WINDSPEED_ADDER_ERROR
+                self.brokenSensorData = [brokenSensorID, theta, windspeedMultiplier, windspeedAdder]
+
+            actualAngle = self.get_wind_measurement(data, self.brokenSensorData[0], "angle")
+            actualSpeed = self.get_wind_measurement(data, self.brokenSensorData[0], "speed")
```
patrick-5546, line 191: Can combine this into one function call (return a tuple)
```
actualAngle, actualSpeed = self.get_wind_measurement(data, self.brokenSensorData[0])
```

python/MOCK_sensors.py
```diff
@@ -122,8 +131,151 @@ def publish(self):
 
         if rospy.get_param('sensor_noise', default=False):
             data = self.add_noise(data)
+
+        brokenWindSensorStatus = rospy.get_param('broken_wind_sensors', default="not_broken")
+        if brokenWindSensorStatus != "not_broken":
+            rospy.logwarn("Broken Wind Sensor Status: " + brokenWindSensorStatus)
```
patrick-5546, line 137: Can `return` at the end of this

python/MOCK_sensors.py
```diff
@@ -122,8 +131,151 @@ def publish(self):
...
+            # brokenSensorData is a list containing: brokenSensorID, the constant angle, the constant direction.
+
+            if(self.brokenSensorData == "uninitialized"):
+                randomSeed = get_rosparam_or_default_if_invalid('random_seed', default=None, rosparam_type_cast=str)
+                random.seed(randomSeed)
+
+                brokenSensorID = random.randint(1, 3)
+                brokenSensorAngle_deg = random.random() * 360
+                brokenSensorSpeed_kts = random.random() * MAX_BROKEN_WINDSPEED_KTS
+                self.brokenSensorData = [brokenSensorID, brokenSensorAngle_deg, brokenSensorSpeed_kts]
```
patrick-5546, line 166: - Document `self.brokenSensorData` contents for each case
- Might be clearer to make it a dictionary

python/MOCK_sensors.py
```diff
@@ -122,8 +131,151 @@ def publish(self):
...
+
+                brokenSensorID = random.randint(1, 3)
+                loopsUntilBroken = random.randint(MIN_LOOPS_UNTIL_BROKEN, MAX_LOOPS_UNTIL_BROKEN)
+                self.brokenSensorData = [brokenSensorID, loopsUntilBroken, 0, 0, 0]
+                # elements at index 2, 3, and 4 are loop counter, last angle, and last speed
+
+            if self.brokenSensorData[2] < self.brokenSensorData[1]:
+                self.brokenSensorData[2] += 1
+            elif self.brokenSensorData[2] == self.brokenSensorData[1]:
+                self.brokenSensorData[2] += 1
```
patrick-5546, line 225: Is this if else statement needed? Both cases are the same

patrick-5546 (commented):
Some small changes that would make the code cleaner

python/MOCK_sensors.py
patrick-5546: Access `random.randint(1, 3)` with `self.brokenSensorData['brokenSensorID']`

python/MOCK_sensors.py
```diff
@@ -122,8 +130,148 @@ def publish(self):
...
+    def breakWindSensors(self, data, brokenWindSensorStatus):
+        # Corrupts wind sensor data depending on the setting specified by string brokenSensorStatus
+
+        if (brokenWindSensorStatus == "stuck_at_zero"):
+            # This setting chooses a wind sensor which reads an angle of 0 and speed 0
+
+            if(self.brokenSensorData is None):
+                randomSeed = get_rosparam_or_default_if_invalid('random_seed', default=None, rosparam_type_cast=str)
+                random.seed(randomSeed)
+                self.brokenSensorData = random.randint(1, 3)
```
patrick-5546, line 150: Since these lines are executed in every branch, they can be moved to before the if-else statements

patrick-5546: Instead of directly setting `self.brokenSensorData` when moved before the if-else statements, can create a new variable (e.g. `brokenSensorID`)

python/MOCK_sensors.py
```diff
@@ -122,8 +130,148 @@ def publish(self):
...
+                                         'windspeedMultiplier': multiplier,
+                                         'windspeedAdder': (2*random.random() - 1) * MAX_WINDSPEED_ADDER_ERROR_KNOTS}
+
+            actualAngle, actualSpeed = self.get_wind_measurement(data, self.brokenSensorData['brokenSensorID'])
+
+            newAngle = actualAngle + self.brokenSensorData['theta']
+            newSpeed = max(0, self.brokenSensorData['windspeedMultiplier'] * actualSpeed
+                           + self.brokenSensorData['windspeedAdder'])
+
+            # Bound the newAngle to [-180, 360) to improve interpretability
```
patrick-5546, line 206: Should it be [-180, 180)?

python/MOCK_sensors.py
```diff
@@ -122,8 +130,148 @@ def publish(self):
...
+            # theta is within [-MAX_ANGLE_ERROR_DEG, MAX_ANGLE_ERROR_DEG]
+
+            if(self.brokenSensorData is None):
+                randomSeed = get_rosparam_or_default_if_invalid('random_seed', default=None, rosparam_type_cast=str)
+                random.seed(randomSeed)
+
+                if random.random < 0.5:
+                    multiplier = 1 - random.random() * (1 - 1 / (MAX_WINDSPEED_MULTIPLIER_ERROR))
+                else:
+                    multiplier = 1 + random.random() * (MAX_WINDSPEED_MULTIPLIER_ERROR - 1)
```
patrick-5546, line 193: Let's discuss this

python/MOCK_sensors.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 import numpy as np
 import rospy
+import random
```
patrick-5546, line 4: Let's use the `numpy` equivalents of the `random` functions that you used to save an import

patrick-5546 (changes requested):
Great work so far, very clean and clear code. Made a couple suggestions

python/MOCK_sensors.py
```diff
@@ -122,8 +130,148 @@ def publish(self):
...
+
+                self.brokenSensorData = {'brokenSensorID': random.randint(1, 3),
+                                         'theta': (2*random.random() - 1) * MAX_ANGLE_ERROR_DEG,
+                                         'windspeedMultiplier': multiplier,
+                                         'windspeedAdder': (2*random.random() - 1) * MAX_WINDSPEED_ADDER_ERROR_KNOTS}
+
+            actualAngle, actualSpeed = self.get_wind_measurement(data, self.brokenSensorData['brokenSensorID'])
+
+            newAngle = actualAngle + self.brokenSensorData['theta']
+            newSpeed = max(0, self.brokenSensorData['windspeedMultiplier'] * actualSpeed
```
patrick-5546, line 203: Can absolute value with `fabs()`

jamenkaye: What was the issue with the built in function abs()? abs() seems to work as expected for floats as far as I can tell

patrick-5546: The output type of abs is the same as its input, while for fabs it is always a float
```
abs(<int>) = <int>
fabs(<int>) = <float>
abs(<float>) = fabs(<float>) = <float>
```

Thus, if the input data is guaranteed to be a float, we can use `abs()`.

jamenkaye: Ok that makes sense. The arguments come from expressions containing `numpy.random.ranf()` so I don't think it should matter, but I changed it to `np.fabs()` just in case.

python/MOCK_sensors.py
jamenkaye: I wasn't satisfied with how I explained this code earlier, so I thought I'd give it another try after looking it over, and work through an example to check myself on my reasoning. Here is an example to show the behaviour I'm going for:

Suppose `MAX_WINDSPEED_MULTIPLIER_ERROR` is set to 1.2. I want the multiplier to be between 1/1.2 = 0.833 and 1.2. The first case is `1 - (a number between 0 and 1 - 1/1.2 = 0.167)`, which gives us a multiplier between 0.833 and 1. The second case is `1 + (a number between 0 and 0.2)`, which gives us a multiplier between 1 and 1.2 as desired.

It would have been more simple to read if the second case just said `muliplier = random.random()*MAX_WINDSPEED_MULTIPLIER_ERROR`, but that leaves the possibility of `random.random()` being close to zero which would lead to the multiplier being outside of the desired range from 0.833 to 1.2.

I could have instead chosen to create a multiplier between `1 - MAX_WINDSPEED_MULTIPLIER_ERROR` and `1 + MAX_WINDSPEED_MULTIPLIER_ERROR`. However, if `MAX_WINDSPEED_MULTIPLIER_ERROR` were set to 1, this would create a multiplier between 0 and 2, and reading 0 all the time seems like a much more severe error than reading double the true windspeed.

(also I realised I was running into issues with integer division, so I changed the 1's to 1.0's)

python/MOCK_sensors.py
```diff
@@ -178,30 +177,32 @@ def breakWindSensors(self, data, brokenWindSensorStatus):
...
 
-                if random.random < 0.5:
-                    multiplier = 1 - random.random() * (1 - 1 / (MAX_WINDSPEED_MULTIPLIER_ERROR))
+                if np.random.randint(1, high=3) < 2:
+                    # multiplier will be between 1/MAX_WINDSPEED_MULTIPLIER_ERROR and 1
+                    multiplier = 1.0 - np.random.ranf() * (1.0 - 1.0 / MAX_WINDSPEED_MULTIPLIER_ERROR)
                 else:
-                    multiplier = 1 + random.random() * (MAX_WINDSPEED_MULTIPLIER_ERROR - 1)
+                    # multiplier will be between 1 and MAX_WINDSPEED_MULTIPLIER_ERROR
+                    multiplier = 1.0 + np.random.ranf() * (MAX_WINDSPEED_MULTIPLIER_ERROR - 1.0)
```
patrick-5546, line 193: Great, clean way of fixing the integer division problem :)

patrick-5546 (approved):
Great work on this PR, excited to revisit this once we get some more sensor data

python/MOCK_sensors.py
patrick-5546: Can you post this in a new comment so that it is easier to find? thanks

python/MOCK_sensors.py
```diff
@@ -185,11 +190,11 @@ def breakWindSensors(self, data, brokenWindSensorStatus):
...
 
-                if np.random.randint(1, high=3) < 2:
-                    # multiplier will be between 1/MAX_WINDSPEED_MULTIPLIER_ERROR and 1
+                if np.random.randint(0, 2):
+                    # 50% chance multiplier will be between 1/MAX_WINDSPEED_MULTIPLIER_ERROR and 1
                     multiplier = 1.0 - np.random.ranf() * (1.0 - 1.0 / MAX_WINDSPEED_MULTIPLIER_ERROR)
                 else:
-                    # multiplier will be between 1 and MAX_WINDSPEED_MULTIPLIER_ERROR
+                    # 50% chance multiplier will be between 1 and MAX_WINDSPEED_MULTIPLIER_ERROR
                     multiplier = 1.0 + np.random.ranf() * (MAX_WINDSPEED_MULTIPLIER_ERROR - 1.0)
```
jamenkaye, line 198: Here is an example to show the behaviour I'm going for:

Suppose `MAX_WINDSPEED_MULTIPLIER_ERROR` is set to 1.2. I want the multiplier to be between 1/1.2 = 0.833 and 1.2. The first case is `1 - (a number between 0 and 1 - 1/1.2 = 0.167)`, which gives us a multiplier between 0.833 and 1. The second case is `1 + (a number between 0 and 0.2)`, which gives us a multiplier between 1 and 1.2. Together, these two possibilities give us the multiplier between 0.833 and 1.2 as desired.

It would have been more simple to read if the second case just said `muliplier = random.random()*MAX_WINDSPEED_MULTIPLIER_ERROR`, but that leaves the possibility of `random.random()` being close to zero which would lead to the multiplier being outside of the desired range from 0.833 to 1.2.

I could have instead chosen to create a multiplier between `1 - MAX_WINDSPEED_MULTIPLIER_ERROR` and `1 + MAX_WINDSPEED_MULTIPLIER_ERROR`. However, if `MAX_WINDSPEED_MULTIPLIER_ERROR` were set to 1, this would create a multiplier between 0 and 2, and reading 0 all the time seems like a much more severe error than reading double the true windspeed.

Issue #265, opened 2021-11-27
Take into account current for mock AIS boats


patrick-5546:
Use track made good instead of true heading in the mock AIS boats
- Mock AIS boat moving logic found [here](https://github.com/UBCSailbot/local-pathfinding/blob/master/python/MOCK_AIS.py#L81)
-  Copy moving logic from [here](https://github.com/UBCSailbot/local-pathfinding/blob/master/python/MOCK_sensors.py#L48) once #267 is merged in

patrick-5546:
Speed from exactAIS already incorporates ocean current, so no need to incorporate ocean current into mock AIS boats

tylerlum:
TODO:
* Add track made good and heading for the AISShip, and then properly model this. (eg. should the wedges that represent other ships be in heading or track made good dir? Does it matter?)

PR #272, opened 2021-12-06
Separate track made good and heading

### LINKED ISSUE #267: Break apart track_made_good and true_heading
Right now, our code treats `track_made_good` and `true_heading` as identical. However, this is not always the case. We need to make changes in the code to properly model and handle this difference.

`track_made_good` is purely the direction that the boat is moving, which is independent of our heading. This is calculated as some type of changing in GPS lat/lon position over time.

`true_heading` is the direction that the boat is pointing towards.

From discussion with the team, we believe that, in an environment with 0 ocean current, the sailboat `track_made_good` = `true_heading` is a reasonable approximation. The only time this changes is when we make a sharp turn, but once the boat picks up speed, the "sideways" movement is primarily gone.

Thus, almost all of the difference between `track_made_good` and `true_heading`  is in the ocean current.

patrick-5546:
Resolves #267

### DIFF
--- a/python/MOCK_sensors.py
+++ b/python/MOCK_sensors.py
@@ -31,7 +31,9 @@ class MOCK_SensorManager:
         self.lat = startLat
         self.lon = startLon
         self.headingDegrees = startHeadingDegrees
-        self.speedKmph = startSpeedKmph
+        self.trackMadeGoodDegrees = startHeadingDegrees  # initialize to boat frame direction (correct when current = 0)
+        self.speedKmphBoatFrame = startSpeedKmph
+        self.speedKmphGlobalFrame = startSpeedKmph  # initialize to boat frame speed (correct when current = 0)
         self.globalWindSpeedKmph = startGlobalWindSpeedKmph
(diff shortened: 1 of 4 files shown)


patrick-5546:
I noticed that the boat often sails up current in this run:

```sh
roslaunch local_pathfinding all.launch speedup:=100 ocean_current_speed:=5
```

Is this to be expected? Does our pathfinding cost function take into account ocean current?

patrick-5546:
#### Todo

- [ ] Add a field for track made good in the GPS msg
- [ ] Publish track made good in ros_interface.py

tylerlum:
Future PR: Consider AIS boats' track_made_good vs heading (which one to use?)

patrick-5546:
Once this is merged in to master, the other CI tests should pass

jamenkaye (approved):
It worked on my computer :)

patrick-5546:
> Future PR: Consider AIS boats' track_made_good vs heading (which one to use?)

Related issue: #265

PR #276, opened 2022-02-04
Log filtering logic to wandb

patrick-5546:
Logs a variety of statistics and filtering methods to Wandb so that we can view and compare filters live during dry land testing. See the `windSensorFilters()` docstring for what has been implemented so far, and ideas for more exotic filtering methods. The graphs in [sailbot-default-project](https://wandb.ai/ubcsailbot/sailbot-default-project) have been updated accordingly.

@hhenry01 no need to checkout to this branch if you are only testing 1 wind sensor. Works when all 3 sensors are running, but will need to modify the `WIND_SPEED_FIELDS` and `WIND_DIRECTION_FIELDS` global variables to get it working when 2 sensors are running.

https://github.com/UBCSailbot/local-pathfinding/blob/a62360d57d383a364f611e9cd1cb25b3bf9b328c/python/wandb_log.py#L31-L32

#### Notes
- I arbitrarily chose the value for the `WIND_SPEED_THRESHOLD` global variable, so that can be finetuned
- Idea: time average as a function of wind speed: higher speed -> shorter average
- Consider warning when sensor values are far apart
- Once GPS AIS is working, look into GPS filter options
    - Average
- I created a Wiki page for Wandb, feel free to add to it: https://github.com/UBCSailbot/local-pathfinding/wiki/Wandb

#### Potential Resources
- Kalman filter
    - https://towardsdatascience.com/kalman-filter-an-algorithm-for-making-sense-from-the-insights-of-various-sensors-fused-together-ddf67597f35e
    - https://www.researchgate.net/publication/267705230_Wind_direction_and_speed_forecasting_optimization_algorithm_for_ship_based_on_time_series_and_Kalman_filter
- Moving average, exponentially weighted moving average: https://hackaday.com/2019/09/06/sensor-filters-for-coders/

### DIFF
--- a/python/wandb_log.py
+++ b/python/wandb_log.py
@@ -2,6 +2,7 @@
 import wandb
 import rospy
 import os
+import numpy as np
 import time
 from log_closest_obstacle import LogClosestObstacle
 from path_storer import PathStorer
@@ -21,9 +22,15 @@ SCREENSHOT = False
 LOG_SENSORS = True
 
 
-# Globals for subscribring
+# Globals for subscribing
 sensor_data = None
 targetGlobalLatLon = None
+
+
+# Globals for wind sensor filtering
+WIND_SPEED_FIELDS = ['wind_sensor_{}_speed_knots'.format(i) for i in range(1, 4)]
+WIND_DIRECTION_FIELDS = ['wind_sensor_{}_angle_degrees'.format(i) for i in range(1, 4)]
+WIND_SPEED_THRESHOLD = 2.0  # wind direction is set to 0 if its speed is below the threshold
 
 
 def sensorsCallback(data):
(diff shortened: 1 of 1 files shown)


patrick-5546:
Does it matter what the wind direction is set to when the wind speed is below the threshold?

Pidud:
https://core.ac.uk/download/pdf/79618904.pdf
Here's the paper I mentioned earlier with extra info on wind sensor filtering, might be something useful here

python/wandb_log.py
```diff
@@ -21,11 +22,17 @@
...
 
-# Globals for subscribring
+# Globals for subscribing
 sensor_data = None
 targetGlobalLatLon = None
 
 
+# Globals for wind sensor filtering
+WIND_SPEED_FIELDS = ['wind_sensor_{}_speed_knots'.format(i) for i in range(1, 4)]
+WIND_DIRECTION_FIELDS = ['wind_sensor_{}_angle_degrees'.format(i) for i in range(1, 4)]
```
patrick-5546, line 32: These values are hardcoded, need to change if not all 3 sensors are running

patrick-5546:
> https://core.ac.uk/download/pdf/79618904.pdf Here's the paper I mentioned earlier with extra info on wind sensor filtering, might be something useful here

Notes from the paper (Chapter 4 - Collecting and Filtering Data):
- Explore the single-pass procedure to compute a mean wind direction
- Moving average filters introduce a delay between the raw measurement and final averaged value, which heavily influences the performance of their implicit tack controller; is this relevant to us?
- Their GPS signals sometimes drops for up to 10 seconds, which would affect our global wind calculations; check if that is the case for us

ubcsailbotsoftware:
> Moving average filters introduce a delay between the raw measurement and final averaged value, which heavily influences the performance of their implicit tack controller; is this relevant to us?

@UBCSailbot/controller can you read over what the implicit tack controller does (Section 6.2) to confirm?

patrick-5546:
> Moving average filters introduce a delay between the raw measurement and final averaged value, which heavily influences the performance of their implicit tack controller; is this relevant to us?

@UBCSailbot/controller can you read over what the implicit tack controller does (Section 6.2) to confirm?

Pidud:
> 

Reading over the section on implicit tack, I think the filtering might be an issue depending on the size of the window we use for filtering. If its short enough it could be okay but a longer one will cause overshooting on tacks but testing would be required to actually figure out if it will be an issue. In the paper they fix the issue by allowing the controller to shorten the averaging window, could be a solution we implement if it becomes an issue.

patrick-5546:
* Closing this testing branch, used to create https://github.com/UBCSailbot/local-pathfinding/pull/283 which has been merged

PR #277, opened 2022-02-27
Handle Low Wind Speed

### LINKED ISSUE #270: Handle global wind speed is 0
When global wind speed is 0, then the direction is measningless. However, we look at the direction wrt the boat's path direction to decide if we are upwind/downwind. THEREFORE, we should have a check that says no upwind/downwind if global wind speed is 0.

jamenkaye:
Publishes a boolean `lowWindConditions` to the `globalWind` message that will be used by Controls to use or stop using the low power controller.

Todo:

- [x] Finetune threshold based on real sensor data
- [x] Decide how to use `lowWindConditions` in the main loop
- [x] Implement using `lowWindConditions` to create local paths less often without slowing down main loop (every ~10s)
- [x] Use multiple thresholds so that we don't continuously switch between using and not using the low power controller
- [x] Push the changes to the sailbot-msg repo: https://github.com/UBCSailbot/sailbot-msg/pull/11
- [x] Low power controller pushed to master branch of controller repo: https://github.com/UBCSailbot/boat_controller/pull/21


jamenkaye:
|  Apparent Wind  |  Boat GPS speed  |  Ideal Outcome  |  Outcome expected from code | 
|-----------------------|--------------------------|----------------------|------------------------------------------|
|  high                   |  any                       |  Sail. If we can feel lots of wind, then sails will work regardless of if the boat is currently moving, or if there is current  |  Exits Low Wind Mode if measured wind is high enough|
|  low                    |  High                      | If there's no current, sail. We could be sailing down wind, where apparent wind is low but we still want to sail. If there IS current, and its the same vector as wind, and we are trying to sail downwind, then apparent wind would be low and GPS speed would be non-zero. In this case, we would want to not sail. | Exits low wind mode if boat speed picks up. If there is current, this could be the "wrong" decision. |
| low                     | low                         | If there's no current, don't sail. If there IS current, and we are trying to sail downwind and upcurrent, then we should sail because at least we would hold steady in current instead of drifting backwards | Enters LWM if apparent wind is low and gps speed is low. If there is current, this could be the "wrong" decision. |

### Comments:

- In the case where apparent wind and GPS speed are low, but we're sailing downwind and upcurrent, I think its acceptable to enter LWM anyway. At least LWM is still sailing, just less actively. Interestingly, if we give up and drift with the current, this pushes us upwind, which would increase the apparent wind speed and GPS speed. This would lead to cycles of turning off LWM, sailing in a way that decreases GPS speed and apparent wind, entering LWM, then increasing windspeed and GPS speed in the wrong direction. Since GPS speed and measured windspeed are both zero, true wind will also be zero, so using true wind instead wouldn't help anyways.
- In the case where current and true wind are aligned and we are trying to sail downwind, it is worrying that we would be trying to sail despite having no wind relative to the water. I believe Control uses apparent wind, and since there is no apparent wind, the wind angle is noisy, so they would waste power with rudder and sail movements. 

[comments still in progress]

python/main_loop.py
```diff
@@ -9,18 +9,24 @@
 import Path
 import time
 import matplotlib.pyplot as plt
+from sailbot_msg.msg import globalWind
 
 # Constants
 MAIN_LOOP_PERIOD_SECONDS = 0.5
 SMALL_TURN_DEGREES = 10
 LOGGED_LATLONS_PER_LINE = 3
+LOW_WIND_LOOPS_PER_PATH_UPDATE = 20
```
jamenkaye, line 18: While in lowWindConditions, this is the minimum number of loops until the local path can be updated for optimisation reasons.

python/main_loop.py
```diff
@@ -125,20 +143,28 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
...
+            lowWindPermitsUpdate = True
+
         newGlobalPathReceived = sailbot.newGlobalPathReceived
         reachedEndOfLocalPath = localPath.reachedEnd()
         pathNotReachGoal = not localPath.reachesGoalLatlon(state.globalWaypoint)
-        mustUpdateLocalPath = (hasUpwindOrDownwindOnPath or hasObstacleOnPath or isGlobalWaypointReached
+        mustUpdateLocalPath = ((hasUpwindOrDownwindOnPath and lowWindPermitsUpdate)
+                               or hasObstacleOnPath or isGlobalWaypointReached
                                or newGlobalPathReceived or reachedEndOfLocalPath or pathNotReachGoal
                                or localPathUpdateForced)
```
jamenkaye, line 159: The idea here is that we only update the local path if the current path is "invalid" for reasons that actually matter in light wind. For example, I think we want to make sure the path is updated normally when we reach a global waypoint so that we at least have a plan for where we want to go in the future. However, upwind/downwind sailing are meaningless in light wind, so this is not a "valid" reason to update the local path.

python/main_loop.py
```diff
@@ -168,8 +194,8 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
             isLocalWaypointReached = localPath.nextWaypointReached(state.position)
             isTimeLimitExceeded = utils.timeLimitExceeded(lastTimePathCreated, rospy.get_param('speedup', default=1.0))
 
-            generateNewLocalPathToCompare = (costTooHigh or isLocalWaypointReached or isTimeLimitExceeded
-                                             or localPathUpdateRequested)
+            generateNewLocalPathToCompare = ((costTooHigh and lowWindPermitsUpdate) or isLocalWaypointReached
+                                             or isTimeLimitExceeded or localPathUpdateRequested)
```
jamenkaye, line 198: I think the same reasoning should apply here as for the condition on line 156.

python/main_loop.py
```diff
@@ -125,20 +143,28 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
         isGlobalWaypointReached = (localPath.waypointReached(state.position, previousGlobalWaypoint, nextGlobalWaypoint,
                                                              True) or isLastWaypointReached)
 
+        if(lowWindConditions):
+            # ensure we havent updated local path recently before permitting
+            lowWindPermitsUpdate = (createNewLocalPathAttempts >= LOW_WIND_LOOPS_PER_PATH_UPDATE)
+            createNewLocalPathAttempts += 1
+        else:
+            lowWindPermitsUpdate = True
```
patrick-5546, line 151: We only call `createNewLocalPath()` when `mustUpdateLocalPath` or `generateNewLocalPathToCompare` is `True`, so would it make sense for the incrementing of `createNewLocalPathAttempts` to depend on these? Or should we stick with the number of loops approach?

Also, is there a reason you didn't reset `createNewLocalPathAttempts` when `lowWindconditions` is `False`?

patrick-5546 (commented):
Great work on this PR, I can see that you have put a lot of thought into the behavior of the `lowWindConditions` boolean, your reasoning makes sense to me. Can you create a new wiki page with your table and explanation? I think it would be helpful to refer to moving forwards.

Also, what have you done to test that your functionality is correct?

patrick-5546:
@UBCSailbot/controller what is the status of low power mode?

jamenkaye:
Testing notes:

- ** Need to change the start boat speed in MOCK_sensors.py line 19 for testing
- Issue: path with angle in it doesn't straighten to direct path on beam reach, even when not in low wind mode. Does the master branch behave like this? Turns out yes, so its not an issue with my code. Added a new issue to github.

[still in progress]

python/main_loop.py
jamenkaye: Good points, I think my next commit addresses both.

python/main_loop.py
```diff
@@ -205,6 +231,11 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
                 else:
                     rospy.loginfo("Keeping old local path")
 
+            elif ((hasUpwindOrDownwindOnPath or costTooHigh)
+                  and (createNewLocalPathAttempts <= CREATE_NEW_LOCAL_PATH_ATTEPT_THRESH)):
+                # Increase the counter if we did NOT update local path, but there WAS an "unimportant" reason to update.
+                createNewLocalPathAttempts += 1
+
```
jamenkaye, line 238: The only time we want to increment `createNewLocalPathAttempts` is when there is at least one reason to update the local path, none of the reasons(s) are "important", and so the attempt to create a new local path was blocked by `lowWindPermitsUpdate`.

If we end up checking the `elif` condition, it means that `mustUpdateLocalPath` was false and that `generateNewLocalPathToCompare` was also false. This already means that there were no "important" reasons to update, and that any "unimportant" reasons that may exist were blocked by `lowWindPermitsUpdate`. The last check we need to do before incrementing `createNewLocalPathAttempts` is to make sure that there WAS an "unimportant" reason, otherwise there was no actual attempt that got blocked.

I also made `createNewLocalPathAttempts` stop updating once it gets big enough, so we don't end up storing a large number if we go a long time before `createNewLocalPathAttempts` resets.

jamenkaye:
lowWindConditions is now its own rostopic, called /lowWindConditions. 

I tested the changes a bit and it seems like everything works fine. So far, I've tested changing the global windspeed with rosparam set, and then using rostopic list to observe /lowWindConditions respond correctly. I still want to spend more time testing how local_pathfinding behaves when lowWindConditions are true.

This topic just contains a boolean, so I don't think there needs to be any mention of Low Wind stuff in the msg repo. I've updated the lowWindConditions branch of the msg repo accordingly. I assume the CircleCI tests I've failed are because the msg repo isnt up-to-date for this code, but let me know if you find any other issues.

python/main_loop.py
patrick-5546: Makes sense!

python/main_loop.py
patrick-5546: > I also made createNewLocalPathAttempts stop updating once it gets big enough, so we don't end up storing a large number if we go a long time before createNewLocalPathAttempts resets.

Good thinking

python/main_loop.py
patrick-5546: It's difficult to say without real world data, can you add this to our testing plan on Confluence? But good work in making it easy to change.

python/main_loop.py
```diff
@@ -60,6 +70,10 @@ def createNewLocalPath(sailbot, maxAllowableRuntimeSeconds, desiredHeadingPublis
         rospy.logerr("Could not find a new localPath within {} seconds. Using previous localPath."
                      .format(maxAllowableRuntimeSeconds))
         return prevLocalPath, lastTimePathCreated
+
+    # Only reset createNewLocalPathAttempts if new path is successfully created
+    global createNewLocalPathAttempts
+    createNewLocalPathAttempts = 0
```
patrick-5546, line 76: Something to consider is to reset this every time `createNewLocalPath()` is called. Regardless if a new path is successfully created or not, the most power-intensive function `Path.createPath()` is called

patrick-5546 (approved):
Very clean code, I also appreciated your explanations of key sections. Once you resolve the unresolved conversations feel free to merge

python/main_loop.py
jamenkaye: Good suggestion, I'd like to talk about this in the meeting tomorrow.

I definitely think it needs to be changed from how I have it right now, because it seems like there's a risk that `createNewLocalPathAttempts` doesn't get defined from the start if the initial local path is None.

My concern was that if we don't have a successful new path, and then 'lowWindPermitsUpdate` blocks creating another new path, could we run out of path to follow? Or maybe end up using a really bad path that puts us in danger? I thought it was safer to only reset the counter if the path we already have was "successful", ie, not None.

jamenkaye: We (Patrick and Jamen) discussed this on July 5th 2022, and decided to leave it as it is.

If we want to further improve power consumption, it is probably more effective and safe to increase `LOW_WIND_LOOPS_PER_PATH_UPDATE` instead of changing this. Changing this parameter makes a difference more continuously, without risking that we get stuck with a None `localPath`

python/main_loop.py
jamenkaye: Added the "Pathfinding: tune `LOW_WIND_LOOPS_PER_PATH_UPDATE`" test item for the [2022-07-16 Wet Testing Plan](https://ubcsailbot.atlassian.net/wiki/spaces/ADA2/pages/1728446482/2022-07-16+Wet+Testing+Plan).

jamenkaye:
I want to test a bit more before merging, just to make sure everything works as expected. See the comments above and below.

Test Procedure (some of this has been done already but I want to double-check):

1. Change `START_BOAT_SPEED_KMPH` in MOCK_sensors.py to be low enough that low wind code comes into effect
2. Add a temporary print statement in main_loop.py to see the loop counter's value each loop.
3. Confirm the loop counter:
     a. resets when `lowWIndConditions` becomes `False` then `True` again
     b. resets when a new path is created
     c. doesn't get larger than just large enough to allow a path update
4. Confirm it enters and exits `lowWindConditions` as expected.
     a. Change windspeed during a run
     b. Change boatspeed between runs
     c. ** Recall that `lowWindConditions` is controlled by apparent wind, whereas we only control the true wind in the simulation.
5. Use extra temporary print statements to make sure control flows correctly in main_loop.py.
6. Keep an eye out for any other fishy business.

jamenkaye:
Testing today: 
- had to checkout e247263 to make local pathfinding run properly (commit where Patrick merged master into LowWindThreshold)
- Temporarily changed `START_BOAT_SPEED_KMPH` in MOCK_sensors.py to be low enough that low wind code comes into effect. The default global wind speed is low enough that this is normally all that is needed to enter `lowWindConditions`. 
- `rospy.logwarn("Low wind conditions: {}. New path attempts: {}".format(lowWindConditions, createNewLocalPathAttempts))` was added temporarily to main_loop.py, right after where it sets `lowWindPermitsUpdate`
- Testing: used `rosparam set /global_wind_direction_degrees` to change wind direction so the path had upwind sailing. Observed the `createNewLocalPathAttempts` counter increase to the threshold, then a new path was created. (As desired)
- **I think `CREATE_NEW_LOCAL_PATH_ATTEPT_THRESH` should be increased.** If Raye is sailing into the wind, pathfinding will want to update the path every loop. With the current threshold, it only takes the counter ~10 seconds to get to the threshold, and then the local path is permitted to update. To save power with `lowWindConditions`, we want to update the path fewer times than we would have over the same period (eg, a few minutes) in the same conditions but without `lowWindConditions`. My concern is that with too low a threshold, then all we achieve is waiting a little bit before updating the path. We want a threshold high enough that without `lowWindConditions`, the path would have been updated multiple times in the amount of time it takes for the counter `createNewLocalPathAttempts` to reach the threshold. I think about 50 or so would be good?
- will resume this comment later

jamenkaye:
**I kept testing in detached head state, checked out to e247263. I'm not sure why but I keep getting this error if I try to run local pathfinding in the latest version of this branch**

Screenshot from 2022-07-30 10-10-12

- Discovered a bug: In `lowWindConditionsCallback()`, `data` is a special ROS datatype. I needed to do `data.data` instead to access the normal boolean. Not sure how I never noticed that earlier.
- Forgot, then remembered, that `lowWindConditions` uses apparent wind, so it's a little different from the global wind. Temporarily changed the start boat speed to be almost 0, so the difference would be minimal.
- Also edited a comment in ros_interface.py to reflect that it uses apparent wind.
- Changed global wind with rosparam set command, and it entered and exited `lowWindConditions` as expected.
- Changed the wind angle so that the path would be bad, and observed the counter increase up to the threshold, then a new path was created. (As expected). The loop counter reset to 0 after the new path was created (As expected).
- While the counter was greater than 0, I increased the wind speed so that it exited `lowWindConditions`. The counter reset to 0, as expected.
- Adjusted the boat speed between runs. This affects `lowWindConditions` as expected.

I think this task will be completed once the following changes are made:
- increase `CREATE_NEW_LOCAL_PATH_ATTEPT_THRESH` to ~40
- in `lowWindConditionsCallback()`, change it to `lowWindConditions = data.data`
- In ros_interface.py, reword comments in `updateWindIsLow()` to avoid "Global wind"

patrick-5546:
> **I kept testing in detached head state, checked out to [e247263](https://github.com/UBCSailbot/local-pathfinding/commit/e247263114d09ef0c8c0ee9d51aca446d0ee3133). I'm not sure why but I keep getting this error if I try to run local pathfinding in the latest version of this branch**

The changes to min_voltage were recently pushed to sailbot-msg, I just merged in the corresponding local-pathfinding changes to this branch:
1. `git pull` sailbot-msg master
2. `catkin_make` in the catkin_ws directory

Then you shouldn't encounter this error anymore

patrick-5546:
> * Discovered a bug: In `lowWindConditionsCallback()`, `data` is a special ROS datatype. I needed to do `data.data` instead to access the normal boolean. Not sure how I never noticed that earlier.

This is a good catch, I will check if min_voltage follows the same behavior during today's dry land testing

jamenkaye:
@patrick-5546 Thanks for the help merging in the min_voltage stuff! It seems to have worked. 

I made the changes described in my previous comment and tested a little bit to make sure it still works with the min_voltage changes. I did more thorough testing of the changes relevant to lowWindThreshold previously, so I don't think any more testing is needed. I think its good to go! 

Feel free to have another look if you like, but I don't think that's necessary. Should I go ahead and hit Squash and merge, and close the Issue?

patrick-5546:
Go ahead 👍

PR #288, opened 2022-07-18
Fixed heading convention?

jamenkaye:
The utilities functions are used in several other places, so I was hesitant to to make the change there. Do still think I should fix the issue in utilities?

Also, currently the only adjustment needed is angles from -90 to 0 need to be increased by 360 to get into the new range. Should I handle if angles are smaller than -360 or larger than 360 as well? I don't think this is necessary but let me know what you think.

### DIFF
--- a/python/utilities.py
+++ b/python/utilities.py
@@ -302,12 +302,7 @@ def headingToBearingDegrees(headingDegrees):
     Returns:
        float representing the same angle in bearing coordinates
     '''
-    bearingDegrees = -headingDegrees + 90
-
-    if (bearingDegrees < 0):
-        bearingDegrees += 360
-
-    return bearingDegrees
+    return bound_in_0_to_360(-headingDegrees + 90)
 
 
 def bearingToHeadingDegrees(bearingDegrees):
@@ -324,7 +319,29 @@ def bearingToHeadingDegrees(bearingDegrees):
     Returns:
        float representing the same angle in heading coordinates
     '''
-    return -bearingDegrees + 90
+    return bound_in_0_to_360(-bearingDegrees + 90)
+
+
+def bound_in_0_to_360(angle):
+    ''' Given an angle in degrees, calculates an equivalent angle bounded in [0, 360)
+
+    Args:
+        angle (float or integer): an angle to be bounded, measured in degrees
+
+    Returns:
+        equivalent angle bounded to [0, 360). The return type matches the type of the input angle.
+    '''
+
+    if angle < 0:
+        returnAngle = 360 - (-angle % 360)
+
+        if returnAngle == 360:
+            return 0
+        else:
+            return returnAngle
+
+    else:
+        return angle % 360
 
 
 def vectorAverage(magnitudes, angles, weights=None):
--- a/python/utilities.py
+++ b/python/utilities.py
@@ -11,8 +11,8 @@ from geopy.distance import distance
 from sailbot_msg.msg import latlon
 
 # Location constants
-PORT_RENFREW_LATLON = latlon(48.5, -124.8)
-MAUI_LATLON = latlon(20.0, -156.0)
+PORT_RENFREW_LATLON = latlon(0, 0)
+MAUI_LATLON = latlon(1, -10)
 
 # Constants
 PATH_UPDATE_TIME_LIMIT_SECONDS = 7200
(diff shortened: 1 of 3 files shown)


jamenkaye:
During the meeting today, Patrick advised that I should check the other places the utilities are used, and if possible move the change to the utility function.

jamenkaye:
As far as I could tell from looking at where `headingToBearingDegrees()` is used, it is safe to change the bounding in the function.

Changes made in this commit:
- Moved the code that bounds the angles to the `headingToBearingDegrees()` function in utilities.py
- Added a brief explanation to the function's description

Tests performed:
- changed the start and end latlons of Maui and Port Renfrew to test the boat sailing on angles in all quadrants. `/desired_heading_degrees` behaved appropriately.
- Ran local pathfinding for a few minutes and used rosparam set to change the wind angle to make sure nothing weird happened. Everything seemed normal except I noticed the warning shown in the picture attached. I think this is fine, right?
- Ran local pathfinding with `enable_los:= true`. Everything still seemed normal.

If you don't see anything you'd like me to change here, I think this issue is completely wrapped up and ready to be merged in and closed.

Screenshot from 2022-07-23 11-10-24

patrick-5546 (changes requested):
I did not encounter the warning you did running this branch on my computer. Did you notice anything that may have caused it? Does it occur in your copy of the master branch?

Also, can you update `headingToBearingDegrees()` and `bearingToHeadingDegrees` to return a value in [0, 360) no matter the input? Then if we ever change how we use them in the future, we won't have to revisit this.

patrick-5546:
> * Ran local pathfinding for a few minutes and used rosparam set to change the wind angle to make sure nothing weird happened. Everything seemed normal except I noticed the warning shown in the picture attached. I think this is fine, right?

I did not encounter the warning you did running this branch on my computer. Did you notice anything that may have caused it? Does it occur in the master branch as well?

patrick-5546 (commented):
> * Ran local pathfinding for a few minutes and used rosparam set to change the wind angle to make sure nothing weird happened. Everything seemed normal except I noticed the warning shown in the picture attached. I think this is fine, right?

I did not encounter the warning you did running this branch on my computer. Did you notice anything that may have caused it? Does it occur in the master branch as well?

jamenkaye:
Sure, I feel like making sure its always bounded to [0, 360) is the more thorough way of doing things. I'll make the change in `headingToBearingDegrees()` and `bearingToHeadingDegrees()`. For the bounding portion of code, I'm thinking something like this:

`calculate angle #unbounded angle
if angle < 0:
    return 360 - ((-1*angle) % 360)
else
    return angle % 360`

eg, with -370, this will return 360 - 10 = 350 as desired

jamenkaye:
As for the warning, I'll keep an eye out for it in the future and see if I can figure it out. Luckily it didn't actually seem to make anything run differently. If you aren't even getting it at all, I'm guessing its just something with my computer and not anything we need to worry about.

jamenkaye:
Created a function to bound any angle to [0, 360), and used the function in both `headingToBearingDegrees()` and `bearingToHeadingDegrees`. 

Tests performed: 
- ran local pathfinding for a few minutes and changed the wind angle to make sure it didn't crash.
- tested the function with these tests: 
[Bound angle in 0 to 360 test.txt](https://github.com/UBCSailbot/local-pathfinding/files/9186300/Bound.angle.in.0.to.360.test.txt)

Ok, NOW I think its done and ready to be merged in, unless there are any other changes you'd like me to make haha

python/utilities.py
```diff
@@ -316,7 +319,29 @@ def bearingToHeadingDegrees(bearingDegrees):
...
+
+    Returns:
+        equivalent angle bounded to [0, 360). The return type matches the type of the input angle.
+    '''
+
+    if angle < 0:
+        returnAngle = 360 - (-angle % 360)
+
+        if returnAngle == 360:
+            return 0
```
patrick-5546, line 339: This case never occurs if `angle < 0`

patrick-5546:
> * tested the function with these tests:
>   [Bound angle in 0 to 360 test.txt](https://github.com/UBCSailbot/local-pathfinding/files/9186300/Bound.angle.in.0.to.360.test.txt)

Can add these to the tests that circleci run every commit. They are found in the test/ directory

python/utilities.py
patrick-5546: Nevermind, this occurs for negative multiples of 360

patrick-5546 (approved):
Great work, I just cleaned up the bounding function a bit and added the tests you wrote to our test suite

PR #298, opened 2022-08-17
Update angle bounding and conventions

patrick-5546:
Numerous updates related to angle bounding and conventions that arose during yesterday's dry land testing
1. Replace `boundIn0To360(num)` with `(num) % 360`
2. Bound the exponentially weighted moving average for angles in [0, 360)
3. Convert wind sensor direction conventions to our software's conventions: the wind sensors identify the direction wind is coming from (e.g., 180 is downwind) whereas our software expects the direction wind is going to (e.g., 0 is downwind)

### DIFF
--- a/python/utilities.py
+++ b/python/utilities.py
@@ -360,14 +360,18 @@ def ewma(current_val, past_ewma, weight, is_angle):
         is_angle (bool): True when getting the field is an angle, false otherwise.
 
     Returns:
-        float: The exponentially weighted moving average.
+        float: The exponentially weighted moving average, bounded in [0, 360) if is_angle is true.
     """
     if past_ewma is None:
         return current_val
 
     if is_angle:
         _, current_val = mitsutaVals((past_ewma, current_val))
-    return weight * current_val + (1 - weight) * past_ewma
+
+    ret_val = (weight * current_val) + ((1 - weight) * past_ewma)
+    if is_angle:
+        ret_val %= 360
+    return ret_val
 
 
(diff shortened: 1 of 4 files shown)


python/utilities.py
```diff
@@ -372,14 +366,18 @@ def ewma(current_val, past_ewma, weight, is_angle):
...
 
     Returns:
-        float: The exponentially weighted moving average.
+        float: The exponentially weighted moving average, bounded in [0, 360) if is_angle is true.
     """
     if past_ewma is None:
         return current_val
 
     if is_angle:
         _, current_val = mitsutaVals((past_ewma, current_val))
```
jamenkaye, line 375: Just to make sure, `ewma()` doesn't need to be able to handle angles that are bounded really weirdly right? With weirdly bounded angles, I think there's a potential for delta in `mitsutaVals()` to be outside of (-540, 540) which could yield faulty results.

If it doesn't need to handle weird angles, would it be a good idea to add the angle bounding requirements on current_val and/or past_ewma in the description for the `ewma()` function?

patrick-5546: Nice catch, I will add that to the docstring. These arguments are implicitly bounded in [0, 360) in the current implementation so no code needs to change:
* `current_val` comes from the second element in the tuple returned by `utils.vectorAverage()`
* `past_ewma` comes from the return value of `utils.ewma()`

jamenkaye (approved):
Looks good! Nice work catching the angle flipping thing :)

PR #304, opened 2022-08-28
Ignore AIS ships by ID

patrick-5546:
Ignore an AIS ship by adding its ID to `AIS_SHIPS_IGNORE_LIST` in `python/Sailbot.py`


patrick-5546:
@hhenry01 what you requested

PR #306, opened 2022-08-28
Add the olympic peninsula land mass files

patrick-5546:
Since we are moving our start point from Point Renfrew to Sooke, we need to create a new land mass for the olympic peninsula so we don't generate paths through it

image

### Usage
Add the argument `land_mass_file:=land_masses/olympic_peninsula_50.0m_08_28_2022-13_48_40.csv`

### Implementation
Generated by manually combining the land mass files generated by the following commands
```
./land_mass_creator.py 48.01 48.40256274 -124.818669 -124.053969 -d 5
./land_mass_creator.py 48.01 48.284504 -124.053969 -123.5540145755 -d 50
```


patrick-5546:
@hhenry01

